PR #62, opened 2017-03-27
CVE-2016-1644 and CVE-2011-1809

nns3455:


cves/CVE-2011-1809.yml
```diff
@@ -31,7 +32,9 @@ fixes_vcc_instructions: |
 fixes:
 - :commit: c7621b85536ad0e53f55572d94f84aff23a4bb67
   :note: ''
-vccs: []
+vccs:
```
mesternefeld: Format for this is a bit strange. It should be something like
 - 498b074bd0db2913cf2c9458407c0d340bbcc05e:
      note: note goes here
This should be the same for fixes

cves/CVE-2011-1809.yml
```diff
@@ -1,6 +1,6 @@
 ---
 CVE: CVE-2011-1809
-CWE: 
+CWE: CWE-416
```
mesternefeld: Just put the number

cves/CVE-2011-1809.yml
```diff
@@ -52,9 +54,10 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
+    The code was unit tested and the fix involved improving these tests.
+  code: true
+  fix: true
```
lxv7763: I think you need more description here. You kind of answer the questions with a yes or no, but try to actually explain how the code was tested and what it involved.

PR #67, opened 2017-03-29
CVE-2015-1277 and CVE-2015-1297

BenWoodworth:


samshiffman321:
you should not create a new PR for changes. pushing to your same branch as the original PR will automatically update the PR on github. just FYI

BenWoodworth:
That's good to know, thank you!

cves/CVE-2015-1277.yml
```diff
@@ -15,7 +15,7 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: Use-after-free vulnerability allows remote attackers to cause a DoS due to a lack of validity checks.
```
jlt8213: This is fairly brief and would benefit from a more in depth look into the vulnerability. What service would be affected by the DoS? What is being checked for validity?

cves/CVE-2015-1277.yml
```diff
@@ -76,7 +80,7 @@ discovered:
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
   answer:
```
jlt8213: It does look like there is some information about how this vulnerability was discovered that you could discuss here. A user and self proclaimed hacker by the name of SkyLined claims to have originally found it here: http://berendjanwever.blogspot.com/2015/07/1503a-chrome-uiaxtreeunserialize-use.html

cves/CVE-2015-1297.yml
```diff
@@ -15,7 +15,9 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  The WebRequest API did not properly concider a request's source before
```
pr7031: Typo here - "Consider"

cves/CVE-2015-1297.yml
```diff
@@ -74,11 +78,12 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: |
+    Discovered while answering question on Stack Overflow
```
pr7031: This is true. However, I think it could benefit from some elaboration - did the person who discovered it know it was a bug immediately? Or did they think they were just answering a question on stack overflow, as normal?

cves/CVE-2015-1297.yml
```diff
@@ -174,4 +179,5 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    Introduced in a bug fix 3 years prior. In fixing a similar bug with permissions, this bug was created.
```
pr7031: More elaboration needed here as well - Many bugs are introduced when fixing other bugs. But I don't think that quite covers what mistakes were made..

cves/CVE-2015-1297.yml
```diff
@@ -130,37 +135,37 @@ lessons:
     If you think of another lesson we covered in class that applies here, feel
     free to give it a small name and add one in the same format as these.
   defense_in_depth:
-    applies: 
+    applies: false
     note: 
   least_privilege:
-    applies: 
+    applies: false
```
pr7031: I think the argument could be made here that following the principal of least privilege could have avoided this bug - If implemented correctly, you can bet the source of a request would be considered (because it needs to be in order to determine privileges needed, aka least privilege)

cves/CVE-2015-1277.yml
```diff
@@ -33,10 +33,12 @@ fixes_vcc_instructions: |
...
 - :commit: 8dc07f045c6287da3404fa8e0fd8b1b3960afbfa
-  :note: ''
+  :note: 'Validates data before serializing'
 - :commit: e3b7faf115bcd4fec56b1658e60abe077d77f379
-  :note: ''
-vccs: []
+  :note: 'Fix dangling pointers by properly deleting tree nodes'
+vccs:
+- :commit: e736e81b857a78fd473aa435747db6fb69b8a5ea
+  :note:
```
nns3455: Commits/notes for VCCs and Fixes should be in this format:
```
  - <commit>:
      note: <note>.
```
Use https://github.com/andymeneely/chromium-vulnerabilities/blob/master/cves/CVE-2011-3092.yml for an example.

cves/CVE-2015-1277.yml
ahartnett: I would change up the way how you word the description to some less technical language.

cves/CVE-2015-1277.yml
```diff
@@ -57,9 +59,11 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
```
ahartnett: Creating a unit test, after the bug was fixed is an interesting way of going about development,

cves/CVE-2015-1297.yml
```diff
@@ -74,11 +78,12 @@ discovered:
...
-  contest: 
+  answer: |
+    Discovered while answering question on Stack Overflow
+  date: 7/16/2015
+  automated: false
+  google: false
+  contest: false
 subsystem:
   question: |
     What subsystems was the mistake in?
```
ahartnett: By looking at the bugs path, you can find the subsystem very easily

PR #83, opened 2017-10-23
CVE-2015-3336 and CVE-2014-1568

Ransom1993:
Edit the CVE-2015-3336 file to include description and answers to various quetsions. Along with other small edits of files

### DIFF
--- a/cves/CVE-2012-4929.yml
+++ b/cves/CVE-2012-4929.yml
@@ -1,6 +1,6 @@
 ---
 CVE: CVE-2012-4929
-CWE:
+CWE: 310
 announced: 2012-09-15 14:55:03.187000000 -04:00
 description_instructions: |
   You can get an initial description from the CVE entry on cve.mitre.org. These
--- a/cves/CVE-2012-4929.yml
+++ b/cves/CVE-2012-4929.yml
@@ -1,6 +1,6 @@
 ---
 CVE: CVE-2012-4929
-CWE: 
+CWE:
 announced: 2012-09-15 14:55:03.187000000 -04:00
 description_instructions: |
   You can get an initial description from the CVE entry on cve.mitre.org. These
@@ -15,11 +15,12 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description:
 bounty:
-  date: 
-  amount: 
-  references: []
+  date: Nov 14 2016
+  amount: 5337.0
+  references:
+   - https://bugs.chromium.org/p/chromium/issues/detail?id=139744#c117
 reviews:
 - 10887026
 - 10880081
@@ -31,7 +32,7 @@ reviews:
 - 10874087
 bugs:
 - 139744
-repo: 
+repo:
 fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
@@ -46,7 +47,7 @@ upvotes_instructions: |
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes:
 unit_tested:
   question: |
     Were automated unit tests involved in this vulnerability?
@@ -59,9 +60,9 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer:
+  code:
+  fix:
 discovered:
   question: |
     How was this vulnerability discovered?
@@ -77,11 +78,11 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer:
+  date:
+  automated:
+  google:
+  contest:
 subsystem:
   question: |
     What subsystems was the mistake in?
@@ -89,8 +90,8 @@ subsystem:
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged.
-  answer: 
-  name: 
+  answer:
+  name:
 interesting_commits:
   question: |
     Are there any interesting commits between your VCC(s) and fix(es)?
@@ -99,10 +100,10 @@ interesting_commits:
     interesting in light of the lessons learned from this vulnerability. Any
     emerging themes?
   commits:
-  - commit: 
-    note: 
-  - commit: 
-    note: 
+  - commit:
+    note:
+  - commit:
+    note:
 major_events:
   question: |
     Please record any major events you found in the history of this
@@ -111,12 +112,12 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer:
   events:
-  - name: 
-    date: 
-  - name: 
-    date: 
+  - name:
+    date:
+  - name:
+    date:
 lessons:
   question: |
     Are there any common lessons we have learned from class that apply to this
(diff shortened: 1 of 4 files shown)


xdmds:
Travis tests failed. please make sure the tests run.

cves/CVE-2012-4929.yml
```diff
@@ -15,11 +15,12 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description:
```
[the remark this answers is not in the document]
laura-king: You need to complete the rest of this in general, all sections need to be filled out.

cves/CVE-2014-1568.yml
```diff
@@ -1,6 +1,6 @@
 ---
 CVE: CVE-2014-1568
-CWE: 
+CWE:
```
laura-king: Don't forget both cwes.

cves/CVE-2014-1568.yml
```diff
@@ -15,32 +15,37 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description:
+  Due the length decording of the Abstract Syntax Notation One allowed an attack of the Bleichenbacher's Low-Exponent Attack on the TLS handshake and certificate signature forgery
```
laura-king: This description needs to provide more context for your average reader. Also, needs to be restructured grammatically.

cves/CVE-2014-1568.yml
```diff
@@ -53,9 +58,9 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: No
+  code: No
```
laura-king: Code and Fix should be in True/False format. Provide more context in the answer section.

cves/CVE-2014-1568.yml
```diff
@@ -71,20 +76,20 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: This was found by a person name Brain Smith who first found it in Mozilla, but reported it to Chromium since it was there too.
```
laura-king: Provide more context on how this issue was found, and restructure the sentence to not be as repetitive.

cves/CVE-2014-1568.yml
```diff
@@ -105,12 +110,12 @@ major_events:
...
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer:
   events:
-  - name: 
-    date: 
-  - name: 
-    date: 
+  - name: Large break from updates until bug fix
```
laura-king: How does a break have a date? Is it the date before or after the break?

cves/CVE-2014-1568.yml
```diff
@@ -171,4 +176,4 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: Not fully understanding the bit level of code open up the oppunity for a Bleichenbacher's Low-Exponent Attack
```
laura-king: This needs more content, and more insight/conclusions. Also spelling/grammar issues.

cves/CVE-2015-3336.yml
```diff
@@ -15,31 +15,33 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: If the url consistted of 'file://' then the browser would not ask the user if they are going to allow or deny for fullscreen and mouse lock. This problem allowed remote attackers to cause denial of service since the user didn't not have the chance to not allow the page the access.
```
laura-king: Spelling/grammar issues.

cves/CVE-2015-3336.yml
```diff
@@ -52,9 +54,9 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: The original code did not have a unit test, however the fix did include adding a unit
+  code: Yes
```
laura-king: True/False again

cves/CVE-2015-3336.yml
```diff
@@ -126,38 +127,38 @@ lessons:
     If you think of another lesson we covered in class that applies here, feel
     free to give it a small name and add one in the same format as these.
   defense_in_depth:
-    applies: 
-    note: 
+    applies: true
+    note: Because this is a very low level bug which a higher fix could have covered it
```
laura-king: You need to elaborate more.

cves/CVE-2015-3336.yml
```diff
@@ -104,12 +105,12 @@ major_events:
...
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer:
   events:
-  - name: 
-    date: 
-  - name: 
-    date: 
+  - name: Multiple revert and attempt at solving another problem around this code
```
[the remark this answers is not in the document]
bam9523: The name of the event should be succinct, so I would recommend replacing the content in this line with "Fixed " and then whatever the name of the problem is.

cves/CVE-2015-3336.yml
```diff
@@ -15,31 +15,33 @@ description_instructions: |
...
 fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: 6eca4fd0c63d0b31729e35ad883fb07b27dfca73
   :note: ''
-vccs: []
+vccs:
+- :commit: 2420e5a85e9db9bf493db47315184856d395c2d5
+  :note: 'Re-factor'
```
bam9523: Please include more information about the commit. "Re-factor" could suggest that the claimed VCC was actually a refactoring of code that was already vulnerable, or it could suggest that the developers *intended* the commit to be a refactor when, in fact, it was introducing the vulnerability. Either way, more details should be provided.

cves/CVE-2015-3336.yml
```diff
@@ -15,31 +15,33 @@ description_instructions: |
...
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
   For the second round of reviewing, you will be giving a certain amount of
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes:
```
bam9523: +3 upvotes. If I had not read about the vulnerability in this context with a paired CVE, I would never have thought about it as more than a UX defect. However, "UI disruption" without user consent in the context of a browser can be just as effective for a DoS attack as a website breach. We can glean from this that context is a significant factor to consider when examining application security.

cves/CVE-2014-1568.yml
```diff
@@ -171,4 +176,5 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: The mistakes were not fully understanding the limitations of the ASN.1 which allowed a Bleichenbacher's Low-Exponent Attack to be able to be done while decoding.
```
bam9523: This appears to simply be a restatement of the vulnerability. I would recommend adding more insights about mistakes made. I realize meaningful observations of this type of vulnerability might be difficult due to how technical it is, but I believe a good starting point may be asking why it was spread across so many different applications.

cves/CVE-2014-1568.yml
```diff
@@ -105,12 +110,12 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer:
```
bam9523: I believe you are allowed to skip this section (with an acknowledgement, of course) if no significant events occur between the VCC and the commit fixing the vulnerability. Although I would not personally consider a break in updates to be a major event, determining whether it qualifies as "major" is up to you.

cves/CVE-2015-3336.yml
Ransom1993: I will get back to this, I have read it but it will require me looking deeper into this commit

cves/CVE-2015-3336.yml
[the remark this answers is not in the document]
Ransom1993: I will have to dig deeper into this one to get that answer for you

cves/CVE-2015-3336.yml
Ransom1993: Making note here to add later in code just so it easier to do it:

Refactor of WebsiteSettingsBubbleController to use PermissionMenuModel. This moves PermissionSelector to a place where both the views and cocoa UI could use it to define the menu for each permission. The developer Leng also was looking into moving the OnSitePermissionChanged function too however was unsure since it would be need a weak pointer to WebsiteSettings. There was a deviation for the PermissionSelectionView contains all of the UI while the cocoa only acts as a bridge between the menu model and the UI. There was no bug fix in this refactor.

PR #87, opened 2017-10-24
CVE-2014-3173 and CVE-2015-1224

JKingSE:
CVE-2014-3173 and CVE-2015-1224

### DIFF
--- a/cves/CVE-2014-3173.yml
+++ b/cves/CVE-2014-3173.yml
@@ -104,7 +104,9 @@ interesting_commits:
     Write a brief (under 100 words) description of why you think this commit was
     interesting in light of the lessons learned from this vulnerability. Any
     emerging themes?
-  commits: There are no commits between the VCC and fix
+  commits:
+  - commit: None
+    note: There were no other commits between VCC and fix
   - commit:
     note:
 major_events:
--- a/cves/CVE-2014-3173.yml
+++ b/cves/CVE-2014-3173.yml
@@ -1,6 +1,6 @@
 ---
 CVE: CVE-2014-3173
-CWE: 
+CWE: 908
 announced: 2014-08-26 21:55:05.790000000 -04:00
 description_instructions: |
   You can get an initial description from the CVE entry on cve.mitre.org. These
@@ -15,7 +15,11 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  WebGL (Web Graphics Library) in Google Chrome did not ensure that clear calls interacted properly 
+  with the state of a draw buffer. This could allow a denial of service attack by 
+  reading uninitialized memory, which causes weird states in the code.  Also, it leaked virtual 
+  RAM but it was only from sampled textures so this wasn't much of a concern.
 bounty:
   date: 2014-08-26 10:00:00.000000000 -04:00
   amount: 2000.0
@@ -31,9 +35,11 @@ fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
-- :commit: 3f2277869057d588919cfe4122967e701894db5b
+- :commit: a97737c7aaa80d88a4a69b5106b72303fbde341e
   :note: ''
-vccs: []
+vccs:
+- :commit: ee7579229ff7e9e5ae28bf53aea069251499d7da
+  :note: They were working on a fix for a separate bug involving draw buffers.
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
@@ -54,9 +60,12 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
+    There were unit tests initially involved with the code around the bug, but 
+    nothing for the bug.  After fixing, they wrote a test that ensures that the 
+    fix is called at least once.
+  code: true
+  fix: true
 discovered:
   question: |
     How was this vulnerability discovered?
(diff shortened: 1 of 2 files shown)


cves/CVE-2014-3173.yml
```diff
@@ -15,7 +15,11 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
```
rxh4133: You aren't supposed to mention chrome. Honestly I think this instance is okay, but you might want to change that according to the instructions.

cves/CVE-2014-3173.yml
```diff
@@ -54,9 +60,12 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
```
rxh4133: I'm glad to see that they were testing their fixes. What specifically does the test do? How does it make sure that clear calls are interacting properly?

cves/CVE-2014-3173.yml
```diff
@@ -172,4 +188,16 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    This issue appeared to be a coding and logic mistake.  The potential problems
```
rxh4133: How easy would it be to intentionally cause a DOS attack?  Would this crash chrome? What would the possible outcomes be?

cves/CVE-2014-3173.yml
```diff
@@ -106,10 +117,12 @@ major_events:
...
-  answer: 
+  answer: |
+    They had a sizable addition of code when the bug was introduced. They
+    were patching a previous issue.
   events:
-  - name: 
-    date: 
+  - name: Patch 
+    date: 2014-06-06
   - name:
```
laura-king: Were there any events that were team specific? Any interesting commits or conversation?

cves/CVE-2014-3173.yml
```diff
@@ -172,4 +188,16 @@ mistakes:
...
+  answer: |
+    This issue appeared to be a coding and logic mistake.  The potential problems
+    that were caused by the uninitialized buffer being called were not thought of
+    in the initial implementation of the patch they created.  This caused a buffer
+    overflow.  There was not much of an issue with leaking sensitive data however
+    because it only accessed sampled textures, but there was a potential risk of a
+    denial of service attack do to unwanted states being caused by an exploit.
+
+    Their solution to this bug was not one of the possible mitigations on the CWE
+    page, but still affective.  They just check the buffer to see if it is
```
laura-king: *effective

cves/CVE-2015-1224.yml
```diff
@@ -172,4 +189,14 @@ mistakes:
...
-  answer: 
+  answer: |
+    This vulnerability was a bit of a miscommunication and design mistake.  Initially,
+    they did have support for video streams with alpha buffers larger than the image 
+    buffer, but the library changed a year after the first implementation and
+    introduced this vulnerability.  They didn't account for that and this vulnerability
+    was introduced.
+
+    To fix this, they added a check that threw an error if the alpha buffer was larger
+    than the imaage buffer.  These buffers are heap buffers, so this applies to the
```
laura-king: image*

cves/CVE-2014-3173.yml
Norwal: Thanks for pointing that out to me! I made the change in the file so it doesn't mention chrome.

cves/CVE-2014-3173.yml
Norwal: In this case, the unit test I found wasn't really testing the functionality of the clear calls.  The test only checked to see if their function was called only exactly one time.

cves/CVE-2014-3173.yml
Norwal: The person who reported the bug said that they weren't actually able to create a situation that caused a DOS attack while testing.  Only that the possibility was there.  I imagine that this attack, if implemented correctly though, would freeze someones chrome until they force closed it and got off the page holding the exploit.

cves/CVE-2014-3173.yml
Norwal: There weren't any interesting commits with this bug because the VCC and the fixing commit were right after each other.

cves/CVE-2014-3173.yml
Norwal: Thanks for pointing that out, I'll fix it.

cves/CVE-2015-1224.yml
Norwal: Thanks for pointing that out, I'll fix it.

cves/CVE-2014-3173.yml
```diff
@@ -1,6 +1,6 @@
 ---
```
tommydvelopment: I award this 2 of my upvotes.

cves/CVE-2015-1224.yml
```diff
@@ -1,6 +1,6 @@
 ---
```
tommydvelopment: I award this 3 of my upvotes.

cves/CVE-2015-1224.yml
```diff
@@ -15,7 +15,11 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  Google Chromes VPX Decoder, which decodes images, does not ensure that alpha-plane
```
tommydvelopment: They told us not to mention chrome, but honestly I would have used it here for describing this. Just playing devils advocate. If you choose to keep it, I would change "Chromes" to "Chrome's."

cves/CVE-2015-1224.yml
```diff
@@ -172,4 +189,14 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
```
tommydvelopment: I would also say that this is an issue with maintainability. Do you think the fix will cause other issues long term? Just checking file size seems like it could lead to another unexpected vulnerability later down the line when more things are changed, I can see this winding up in a VCC later down the line

tommydvelopment: I would also say that this is an issue with maintainability. Do you think the fix will cause other issues long term? Just checking file size seems like it could lead to another unexpected vulnerability later down the line when more things are changed, I can see this winding up in a VCC later down the line.

cves/CVE-2014-3173.yml
Norwal: I marked it down.

cves/CVE-2015-1224.yml
Norwal: I marked it down.

cves/CVE-2015-1224.yml
Norwal: I chose to make it more general and not mention chrome.

cves/CVE-2015-1224.yml
Norwal: It is possible that future issues may occur, but I don't think it was a maintainability issue this time around.  They didn't have trouble fixing the vulnerability once it was found.

PR #94, opened 2017-10-25
CVE-2011-3049 and CVE-2015-1299

tfaticone:
CVE-2011-3049 and CVE-2015-1299 by Thomas Faticone in Predeep's section

### DIFF
--- a/cves/CVE-2011-3049.yml
+++ b/cves/CVE-2011-3049.yml
@@ -16,11 +16,11 @@ description_instructions: |
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
 description: |
-  Web requests are made all the time by browsers looking to recieve content,
+  Web requests are made all the time by browsers looking to retrieve content,
   post content, and the like. In this case the API in question allows developers
   to view these requests to take analytics, block them, or modify them. This is
   problematic in the case that a web extension goes rogue. This api allows rogue
-  extensions to block incoming requests. In this case, it allows a rogue extension
+  extensions to block incoming requests. Specifically, it allows a rogue extension
   to block a request calling for the blacklist of web extensions (which could include
   itself).
 bounty:
@@ -40,17 +40,18 @@ fixes_vcc_instructions: |
 fixes:
 - :commit: 22503495dc014e883d1041f6b9d883a1a53f3e55
   :note: |
-    Cleaned up so that URL tested (to make sure its not an a Webstore URL) is without parameters/query statements.'
+    Cleaned up so that URL tested (specficially a Webstore URL) is without parameters/query statements.'
 - :commit: a8ecc56a5fbdbb537ecc9559ec270985d92e3908
   :note: |
-    The fix for the issue. Explicitly checks whether the blacklist url is being accessed and hides it from extensions (to avoid blocking it).'
+    The fix for the issue. Explicitly checks whether the requested url that is being accessed is the update blacklist URL. If so,
+    the code hides it from extensions (to avoid blocking it).
 vccs:
 - :commit: 67351452e1e16cb9b01c609ce185a9e819cf2254
   :note: |
     Originally used to remove the use of ProfileId because it allowed profile
-    to be used improperly. But did not understand that only relying on the profile context
+    to be used improperly. The coder, jam@chromium.org, did not understand that only relying on the profile context
     to decide whether to hide the request from extensions was not enough. As this allowed blocking
-    of requests to update the extension blacklist.
+    of requests that were meant to update the extension blacklist.
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
@@ -98,9 +99,9 @@ discovered:
   answer: |
     The vulnerability was found by someone working with extensions. They were working with
     webrequest.onBeforeRequest, which is an event that is fired just before a request is sent
-    out, and were able to intercept it and block it be looking to see whether the url has blocklist
-    in it. They initially noticed this capability by looking at their logs of onBeforeRequest events.
-    In addition, they also noticed this could be used to block the Web Store.q
+    out, and were able to intercept it the request, check to see whether it was an update to the blacklist,
+     and block it. They initially noticed this capability by looking at their logs of onBeforeRequest events.
+    In addition, they also noticed this could be used to block the Web Store.
   date: 2011-12-26
   automated: false
   google: false
@@ -162,6 +163,7 @@ lessons:
     applies: true
     note: |
       Extensions should not have been given the permissions to block blacklist calls.
+      In my opinion, there should have been a thorough list of non-blockable request.
   frameworks_are_optional:
     applies:
     note:
@@ -211,6 +213,6 @@ mistakes:
     the possibility of blocking important connections through a rogue extension. Furthermore, there seemed to be alot
     of misunderstanding dealing with how the updates to the blacklist were handled. Looking at the bug
     comments, there was a good amount of debate about how the updating of the extensions was
-    initially planned to be a system call (while not using the profile) but ended up using the
+    initially planned to be a system call (meaning not using the profile) but ended up using the
     web request. In retrospect, if they had followed through with the initial design choice, the
     vulnerability would have been avoided.
--- a/cves/CVE-2011-3049.yml
+++ b/cves/CVE-2011-3049.yml
@@ -1,6 +1,6 @@
 ---
 CVE: CVE-2011-3049
-CWE: 
+CWE:
 announced: 2012-03-23 06:55:01.113000000 -04:00
 description_instructions: |
   You can get an initial description from the CVE entry on cve.mitre.org. These
@@ -15,24 +15,42 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  Web requests are made all the time by browsers looking to recieve content,
+  post content, and the like. In this case the API in question allows developers
+  to view these requests to take analytics, block them, or modify them. This is
+  problematic in the case that a web extension goes rogue. This api allows rogue
+  extensions to block incoming requests. In this case, it allows a rogue extension
+  to block a request calling for the blacklist of web extensions (which could include
+  itself).
 bounty:
-  date: 
-  amount: 
+  date:
+  amount:
   references: []
 reviews:
+- 8952021
 - 9120011
 - 9152010
 bugs:
 - 108648
-repo: 
+repo:
 fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: 22503495dc014e883d1041f6b9d883a1a53f3e55
-  :note: ''
-vccs: []
+  :note: |
+    Cleaned up so that URL tested (to make sure its not an a Webstore URL) is without parameters/query statements.'
+- :commit: a8ecc56a5fbdbb537ecc9559ec270985d92e3908
+  :note: |
+    The fix for the issue. Explicitly checks whether the blacklist url is being accessed and hides it from extensions (to avoid blocking it).'
+vccs:
+- :commit: 67351452e1e16cb9b01c609ce185a9e819cf2254
+  :note: |
+    Originally used to remove the use of ProfileId because it allowed profile
+    to be used improperly. But did not understand that only relying on the profile context
+    to decide whether to hide the request from extensions was not enough. As this allowed blocking
+    of requests to update the extension blacklist.
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
@@ -40,7 +58,7 @@ upvotes_instructions: |
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes:
 unit_tested:
   question: |
     Were automated unit tests involved in this vulnerability?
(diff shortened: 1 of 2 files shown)


cves/CVE-2015-1299.yml
```diff
@@ -147,38 +174,38 @@ lessons:
...
   yagni:
-    applies: 
-    note: 
+    applies: true
+    note: There was a problem about making a complex heap structure when it did not need it. (So it was removed)
   complex_inputs:
-    applies: 
-    note: 
+    applies: true
+    note: There was a theory that since they were comparing an 80-bit integer and a 64-bit integer the equality (they assumed) would work did not.
```
lveilleux: I don't know if this vulnerability should be considered a complex input, it looks like this is all behind the scenes and would occur without specific user interaction from the browser. Thoughts?

cves/CVE-2015-1299.yml
```diff
@@ -147,38 +174,38 @@ lessons:
...
+    note:
   least_privilege:
-    applies: 
-    note: 
+    applies:
+    note:
   frameworks_are_optional:
-    applies: 
-    note: 
+    applies:
```
lveilleux: Just a thought, it seems like this may apply here, given that the solution they had was to revamp the entire system, from the original timers thrown into a heap.

cves/CVE-2015-1299.yml
```diff
@@ -191,4 +218,14 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    The vulnerability was a complex issue that never really led to a defined solution. It
+    started with the fact they made the decision to create a heap and use that for a scheduler.
+    This is an issue because it was not needed and added a good amount of complexity which makes
+    security more difficult. In addition, it was in a relatively unpopular area with only a few people
+    being familiar with the section. This led for the bug to exist for quite sometime after it was first found.
```
lveilleux: This led.... reads a weird, did you mean 'allowed'?

cves/CVE-2015-1299.yml
```diff
@@ -15,7 +15,15 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  There was an issue with timers being removed and the reference being used.
+  Timers were set up so that they lived in a heap. The problem with this is there
+  was logic that ended up freeing/removing a timer, due to a bad equality that
+  attempted to compare an 80-bit integer with a 64-bit integer, and
+  the program would attempt to use the reference to the now freed timer.
```
lveilleux: The majority of this paragraph is one running sentence, would you mind splitting it up a bit?

cves/CVE-2011-3049.yml
```diff
@@ -53,9 +72,15 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
+    In this section of code there were unit tests that are run on this file/module.
+    I found ths by going back to the commit where the fix was made an inspecting the same
```
lveilleux: There's a few grammatical & spelling mistakes here; 'ths', 'an'

cves/CVE-2011-3049.yml
```diff
@@ -127,38 +157,44 @@ lessons:
...
+    note:
   environment_variables:
-    applies: 
-    note: 
+    applies:
+    note:
   secure_by_default:
-    applies: 
-    note: 
+    applies:
```
lveilleux: It looks like this could apply here as well given that the updater for the blacklist was not hidden within the inner workings of Chromium, but rather in a place that could be intercepted by malicious extensions.

cves/CVE-2015-1299.yml
```diff
@@ -33,34 +41,41 @@ reviews:
...
+  :note: 'Potential VCC. Introduced a function that stores a variable and returns the 80-bit interger. This is the leading idea of what caused the UAF (use after free).'
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
   For the second round of reviewing, you will be giving a certain amount of
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes:
```
lveilleux: Have some upvotes due to the rather comedic handling of the fix for this issue along with this comment <https://bugs.chromium.org/p/chromium/issues/detail?id=416362#c37>
Upvotes: 4

cves/CVE-2015-1299.yml
```diff
@@ -91,20 +110,24 @@ discovered:
...
     What subsystems was the mistake in?
 
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged.
-  answer: 
-  name: 
+  answer:
+    The Webkit subsystem since that is included in the path. Using the Blink timer.
+  name: Webkit
```
lveilleux: It looks like below on line 157 you mention a rework of the Blink subsystem, but here you say that Webkit was the subsystem the vulnerability was in. Would it be incorrect to say that this was in the Blink subsystem since that's where the 80-bit & 64-bit comparison was done?

cves/CVE-2011-3049.yml
tfaticone: This could completely be applicable depending on how you interpret why the vulnerability was introduced. "Secure by default" is the assumption that the current setup is the most secure way a system can be set up. Knowing that, if the developer intentionally made the updates to the blacklist use web requests then yes, secure by default is an issue because obviously it was not the most secure setup. But in my interpretation, it seemed to be an accident that this happened, they planned to hide the blacklist update from extensions but somewhere between planning and construction it was messed up.

cves/CVE-2015-1299.yml
tfaticone: This is a good question that confused me, doing a little investigation it seems that Blink was a separate subsystem initially but currently it was mixed in with WebKit. I think for reasons of clarity, I'll change this to Blink considering that is where it was when it was found and fixed.

cves/CVE-2015-1299.yml
tfaticone: Hmm that is an interesting thought. I guess my perspective was within the system it would have been considered a complex input but on second thought it definitely is not. Coming from an outsiders perspective (outside of the chromium project) this is definitely not considered an input, let alone a complex input.

cves/CVE-2011-3049.yml
```diff
@@ -71,20 +96,25 @@ discovered:
...
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: |
+    The vulnerability was found by someone working with extensions. They were working with
+    webrequest.onBeforeRequest, which is an event that is fired just before a request is sent
+    out, and were able to intercept it the request, check to see whether it was an update to the blacklist,
+     and block it. They initially noticed this capability by looking at their logs of onBeforeRequest events.
+    In addition, they also noticed this could be used to block the Web Store.
```
amw7010: Who specifically would be blocked from accessing the Web Store?

cves/CVE-2011-3049.yml
```diff
@@ -127,38 +157,44 @@ lessons:
...
-    note: 
+    applies:
+    note:
   least_privilege:
-    applies: 
-    note: 
+    applies: true
+    note: |
+      Extensions should not have been given the permissions to block blacklist calls.
+      In my opinion, there should have been a thorough list of non-blockable request.
```
amw7010: "Request" should be "requests."

cves/CVE-2011-3049.yml
```diff
@@ -127,38 +157,44 @@ lessons:
...
+    note:
   distrust_input:
-    applies: 
-    note: 
+    applies: true
+    note: |
+      The fix involved stunting the ability of the extension to grab hold of requests
+      and cancel them. While this means the fix is a two sided issue, one side was definitely
+      the problem of 'overtrusting' an extensions by allowing it to grab hold of any request. Even
+      one like an update to the blacklist.
```
amw7010: You mention that this was a two-sided issue, and go into detail on one side, but what is the other side?

cves/CVE-2011-3049.yml
```diff
@@ -171,4 +207,12 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    Overall in my opinion there was not enough code review that went into the VCC's merging.
+    While it is difficult to think of such cases, further analysis might have reavealed
```
amw7010: Typo; "reavealed" should be "revealed."

cves/CVE-2015-1299.yml
tfaticone: It is a good idea, if you were to think of the heap as the framework then definitely. This could be a good example of a framework being misused and removed after it was discovered.

cves/CVE-2015-1299.yml
tfaticone: Fixed!

cves/CVE-2011-3049.yml
tfaticone: Good catches, thanks again!

cves/CVE-2011-3049.yml
```diff
@@ -15,32 +15,51 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  Web requests are made all the time by browsers looking to retrieve content,
+  post content, and the like. In this case the API in question allows developers
+  to view these requests to take analytics, block them, or modify them. This is
```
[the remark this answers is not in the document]
tfaticone: I changed the first 'them' I feel as if replacing might sound a bit too redundant.

cves/CVE-2011-3049.yml
```diff
@@ -15,32 +15,51 @@ description_instructions: |
...
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: 22503495dc014e883d1041f6b9d883a1a53f3e55
-  :note: ''
-vccs: []
+  :note: |
+    Cleaned up so that URL tested (specficially a Webstore URL) is without parameters/query statements.'
+- :commit: a8ecc56a5fbdbb537ecc9559ec270985d92e3908
+  :note: |
+    The fix for the issue. Explicitly checks whether the requested url that is being accessed is the update blacklist URL. If so,
```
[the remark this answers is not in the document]
tfaticone: Good catch!

cves/CVE-2011-3049.yml
```diff
@@ -15,32 +15,51 @@ description_instructions: |
...
+    the code hides it from extensions (to avoid blocking it).
+vccs:
+- :commit: 67351452e1e16cb9b01c609ce185a9e819cf2254
+  :note: |
+    Originally used to remove the use of ProfileId because it allowed profile
+    to be used improperly. The coder, jam@chromium.org, did not understand that only relying on the profile context
```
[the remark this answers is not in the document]
tfaticone: Good catch and seems like profiles was undergoing a large change when the chromium developers were trying to remove some dependencies between chrome and the content module. One portion of this issue was to remove the use of profile from content. It seems that this VCC was in the crossfire because the use of profileID allowed profile to get used somewhere within the content module. In accordance with the large project, they removed all uses of profile within content which caused some changes to occur within the vulnerable file. Now this answer is my educated since it seems the overall project was not kept track of well.

cves/CVE-2011-3049.yml
```diff
@@ -71,20 +96,25 @@ discovered:
...
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: |
+    The vulnerability was found by someone working with extensions. They were working with
+    webrequest.onBeforeRequest, which is an event that is fired just before a request is sent
+    out, and were able to intercept it the request, check to see whether it was an update to the blacklist,
+     and block it. They initially noticed this capability by looking at their logs of onBeforeRequest events.
```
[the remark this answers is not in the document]
tfaticone: Seems like the only identification I could get was  adblockf...@gmail.com, which is an abbreviation of a larger email. In addition, it seems like they might have been developing an extensions, potentially an ad blocker (given the name of their email).

cves/CVE-2011-3049.yml
tfaticone: People who have the rogue extension installed would be blocked from accessing the web store.

cves/CVE-2011-3049.yml
tfaticone: Good catch!

cves/CVE-2011-3049.yml
tfaticone: The other side is that there was an issue with how updated to the blacklist were being handled. For sake of specificity, I did not include it in the description of distrust input because it did not belong. Although, I did go into some detail in the mistakes section.

cves/CVE-2015-1299.yml
tfaticone: Whoops you're correct. Thanks for that!

cves/CVE-2011-3049.yml
```diff
@@ -15,32 +15,59 @@ description_instructions: |
...
+    of requests that were meant to update the extension blacklist.
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
   For the second round of reviewing, you will be giving a certain amount of
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes:
```
amw7010: I would like to award 3 of my upvotes to CVE-2011-3049. Getting around any blacklist is interesting on its own, but getting around a blacklist in such a way that an extension is able to get around its *own* blacklist is really intriguing.

PR #103, opened 2017-10-25
CVE-2013-6622 and CVE-2016-1673

bam9523:
# CVE-2013-6622

CVE involves a use-after-free bug resulting from incorrect load delay timing of a media element. Elements were able to send a load event after being destroyed, creating the vulnerability.

# CVE-2016-1673

CVE involves a browser-based XSS vulnerability resulting from lack of adherence to the Same Origin Policy. Requests were able to bypass checks determining whether request navigation should be allowed when they were sent into a frame in a new document.

### DIFF
--- a/cves/CVE-2013-6622.yml
+++ b/cves/CVE-2013-6622.yml
@@ -64,7 +64,7 @@ upvotes_instructions: |
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes: 4
 unit_tested:
   question: |
     Were automated unit tests involved in this vulnerability?
--- a/cves/CVE-2013-6622.yml
+++ b/cves/CVE-2013-6622.yml
@@ -140,7 +140,7 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: Not any major events were seen during the vulnerability.
+  answer: No major events were seen during the vulnerability.
   events:
   - name: 
     date: 
--- a/cves/CVE-2013-6622.yml
+++ b/cves/CVE-2013-6622.yml
@@ -99,7 +99,7 @@ discovered:
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
   answer: |
-    The vulnerability was found by a user on the issue tracking system, who
+    The vulnerability was found by cloudfuzzer on the issue tracking system, who
     appears to have a record of finding bugs (they celebrate their "10th bug
     report" in the report). However, although they did document the stack trace
     and unit test used, they did not document how they came across the
--- a/cves/CVE-2013-6622.yml
+++ b/cves/CVE-2013-6622.yml
@@ -1,6 +1,6 @@
 ---
 CVE: CVE-2013-6622
-CWE: 
+CWE: CWE-416
 announced: 2013-11-13 10:55:04.237000000 -05:00
 description_instructions: |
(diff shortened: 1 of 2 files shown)


cves/CVE-2013-6622.yml
```diff
@@ -72,10 +98,15 @@ discovered:
...
-  answer: 
-  date: 
-  automated: 
-  google: 
+  answer: |
+    The vulnerability was found by a user on the issue tracking system, who
+    appears to have a record of finding bugs (they celebrate their "10th bug
+    report" in the report). However, although they did document the stack trace
+    and unit test used, they did not document how they came across the
+    vulnerability.
```
Ransom1993: Should provide user name of the user who found this bug

cves/CVE-2013-6622.yml
```diff
@@ -15,7 +15,22 @@ description_instructions: |
...
   keep too.
-description: 
+description: |
+  The vulnerability was a "use-after-free" bug in the fork of WebKit used by
+  Chrome. The engine provides the ability to move elements between documents,
+  and frees the memory used by the element in the first document before moving
+  the element to the second document. It was found that, with a certain timing,
+  the freed memory used by the element in the first document could be loaded
+  after it was destroyed, creating a potential denial of service.
+
```
danielroach9: +4 upvotes. I wonder how easy/hard it must have been to get the timing right to expose this vulnerability.

cves/CVE-2013-6622.yml
```diff
@@ -106,7 +140,7 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer: Not any major events were seen during the vulnerability.
```
danielroach9: Should read: "No major events were seen during the vulnerability."

cves/CVE-2013-6622.yml
danielroach9: Don't know if it's only visible on my end but accidentally highlighted a line. My bad lol

cves/CVE-2016-1673.yml
```diff
@@ -32,7 +36,9 @@ fixes_vcc_instructions: |
 fixes:
 - :commit: f93a0e557dd97cc83d8b55953c2f57c5e2dfe07b
   :note: ''
-vccs: []
+vccs:
+  - commit: cff1a5e1c110ad79007bd7bda87cfde6045cc2af
+    note: It appears they introduced the vulnerability while fixing other bugs.
```
danielroach9: Giving some information about what bugs that were trying to fix would be helpful here.

cves/CVE-2013-6622.yml
bam9523: This strange wording is due to the fact that the word "No" is reserved in YAML syntax, resolving to `false`; I was concerned that the CI would reject my submission if I used it. I'll make the suggested change to see if that actually occurs.

bam9523: It appears my concerns were unfounded; the build passed.

cves/CVE-2016-1673.yml
bam9523: Hopefully I was able to explain it to a satisfactory degree. The bug was apparently an OOPIF-handling error ("Out-of-process iframe") where navigation was not correctly dispatched to a remote frame. I felt a full explanation would be too jargony, so I distilled it to the parts relating to the newly-introduced vulnerability.

cves/CVE-2013-6622.yml
bam9523: Addressed; the bug reporter's name is now included.

PR #107, opened 2017-10-25
CVE 2010 1230 and CVE 2015 1226

iataylor:
The second CVE is not done yet and I don't want to give you guys the half completed CVE.

2010 1230 investigation should be complete.

### DIFF
--- a/cves/CVE-2010-1230.yml
+++ b/cves/CVE-2010-1230.yml
@@ -47,7 +47,7 @@ fixes:
   note:
 - commit: 8af4df6bc98ecfb0d742008e5799997ef507a416
   note:
-vccs:
+vccs: []
 - :commit:
   :note:
 - :commit:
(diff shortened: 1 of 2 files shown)


iataylor:
There were no VCC's for the issue. It was a development omission, not a issue that anyone created.

iataylor:
Same thing with interesting commits. Nothing interesting about them. Simple fixes all.

iataylor:
Currently working on trying to get the yaml syntax correct. Apparently the output is giving me the .rb script errors, so it isn't telling me where the yaml syntax error is located. Working on trying to fix it.

iataylor:
So I got to the point that I only get two errors, one per file, both with vccs. I couldn't figure out the format in time (and that's what I spent the majority of my time on). Otherwise, the syntax should be good and the actual research should be complete.

PR #112, opened 2017-10-26
CVE-2015-8480 and CVE-2012-5152

malmgrens4:
Adding CVE-2012-5152.yml

### DIFF
--- a/cves/CVE-2012-5152.yml
+++ b/cves/CVE-2012-5152.yml
@@ -16,7 +16,10 @@ description_instructions: |
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
   
-description: Extra memory was allocated if the playback rate of a video was higher than normal and the player used the skip feature (think youtube). The extra memory allowed for out-of-bounds read access. 
+description: |
+    Extra memory was allocated if the playback rate of a video was higher 
+    than normal and the player used the skip feature (think youtube). 
+    The extra memory allowed for out-of-bounds read access. 
 bounty:
   date:
   amount:
@@ -32,7 +35,9 @@ fixes_vcc_instructions: |
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: 802ecdb9cee0d66fe546bdf24e98150f8f716ad8
-  :note: 'The problem was addressed. A new piece of information was calculated, the step speed, that helped determine the amount of memory that would be allocated'
+  :note: |
+    The problem was addressed. A new piece of information was calculated, the step speed, 
+    that helped determine the amount of memory that would be allocated
 vccs:
 - :commit: e4eccfdae7c8b34056e7b61450b0be30825fc5d2
   :note: ''
@@ -79,7 +84,7 @@ discovered:
   automated: No
   google: Yes
   contest: No
-subsystem: media
+subsystem: 
   question: |
     What subsystems was the mistake in?
 
@@ -95,7 +100,7 @@ interesting_commits:
     Write a brief (under 100 words) description of why you think this commit was
     interesting in light of the lessons learned from this vulnerability. Any
     emerging themes?
-  commits: No
+  commits:
   - commit: 
     note: 
   - commit: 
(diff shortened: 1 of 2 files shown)


tanner-glantz:
Looks like it doesn't pass the Travis CI test. You will need to fix that in order for this PR to be merged.

cves/CVE-2012-5152.yml
```diff
@@ -15,10 +15,11 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+  
+description: Extra memory was allocated if the playback rate of a video was higher than normal and the player used the skip feature (think youtube). The extra memory allowed for out-of-bounds read access.
```
nkk5908: Long answers would be more readable if they spanned multiple lines. A "|" can be written on the first line of the answer and the answer text would start on the second line. A new line can be used when a certain number of characters on a line is reached and each line of the answer would need to be indented by two spaces from the yml tag to follow correct yml syntax.

cves/CVE-2012-5152.yml
```diff
@@ -71,28 +74,28 @@ discovered:
...
-  automated: 
-  google: 
-  contest: 
-subsystem:
+  answer: This was discovered with the use of a fuzzer focused on the seek functionality.
+  date: December 11 2012
+  automated: No
+  google: Yes
+  contest: No
+subsystem: media
```
nkk5908: I believe this is to specify a section of information and answers should only be written next to the answer and name tags below, but I could be wrong. This may be a reason for tests failing because of incorrect yml syntax.

cves/CVE-2012-5152.yml
```diff
@@ -71,28 +74,28 @@ discovered:
...
   question: |
     What subsystems was the mistake in?
 
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged.
-  answer: 
-  name: 
+  answer: media/filters
```
nkk5908: I think text under the answer tags should be complete sentences and specific names should be under the name tags.

cves/CVE-2012-5152.yml
```diff
@@ -71,28 +74,28 @@ discovered:
...
+  name: audio_renderer_algorithm.cc
 interesting_commits:
   question: |
     Are there any interesting commits between your VCC(s) and fix(es)?
 
     Write a brief (under 100 words) description of why you think this commit was
     interesting in light of the lessons learned from this vulnerability. Any
     emerging themes?
-  commits:
+  commits: No
```
nkk5908: There shouldn't be text next to section header tags.

cves/CVE-2012-5152.yml
malmgrens4: Hey, thanks for the tip. Added it to the latest request.

cves/CVE-2012-5152.yml
malmgrens4: This was an issue, thanks for pointing it out.

cves/CVE-2012-5152.yml
malmgrens4: Adjusted this with the latest request, thanks!

cves/CVE-2012-5152.yml
malmgrens4: Fixed this with the latest request.

cves/CVE-2012-5152.yml
```diff
@@ -126,38 +134,38 @@ lessons:
...
-    applies: 
+    applies: false
     note: 
   native_wrappers:
-    applies: 
+    applies: false
     note: 
   distrust_input:
-    applies: 
+    applies: true
```
KadeerAmin: You should justify why you believe that the lesson, distrust input, applies to this vulnerability.

cves/CVE-2015-8480.yml
```diff
@@ -15,9 +15,14 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+     
+description: |
```
KadeerAmin: If a function contains sensitive information, shouldn't it be the responsibility of that function to properly dispose of sensitive information before freeing the memory?

cves/CVE-2015-8480.yml
```diff
@@ -173,4 +180,8 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    Coding mistakes. They neglected to zero initialize the video frame.
```
KadeerAmin: You should try to use full sentences for a formal research assignment.

PR #115, opened 2018-01-31
Worked on 2011-1812 and 2015-1252

siberiax:


siberiax:
@andymeneely Looking into if there is something wrong with my YAML. Wanted you to see as well

andymeneely:
Looks like this isn't still done - I need you to finish this ASAP so you can help me with this project. Students are starting this tomorrow and they'll want feedback.

PR #133, opened 2018-03-19
CVE-2013-0891 and CVE-2014-1732

AndrewDelposen:
Initial pull request for setting up for future commits

### DIFF
--- a/cves/CVE-2013-0891.yml
+++ b/cves/CVE-2013-0891.yml
@@ -11,7 +11,7 @@ curated_instructions: |
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
--- a/cves/CVE-2013-0891.yml
+++ b/cves/CVE-2013-0891.yml
@@ -4,7 +4,7 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: 680
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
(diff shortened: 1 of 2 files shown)


cves/CVE-2013-0891.yml
```diff
@@ -52,9 +55,11 @@ fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
-- :commit: 0ed41ef78a807f5d195b070bce7c433fcf6f5cb3
+- :commit: 811f3cdb481f280a40ca6c3a4df38c78007fc654
```
mesternefeld: Definitely would keep the other commit as well as it is part of getting to the fix

cves/CVE-2013-0891.yml
```diff
@@ -75,9 +80,14 @@ unit_tested:
...
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
+    From looking at the files affected, and the code specifically, there seems to be no
+    automated tests, nor any tests related the files as a whole.
+    
+    The code review and bug tracker did show that the code was being run against the
+    server as continuous test integration, which is why I'm saying "true" for code
```
mesternefeld: Love the explanation here!

cves/CVE-2013-0891.yml
```diff
@@ -52,9 +55,11 @@ fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
-- :commit: 0ed41ef78a807f5d195b070bce7c433fcf6f5cb3
+- :commit: 811f3cdb481f280a40ca6c3a4df38c78007fc654
   :note: ''
-vccs: []
+vccs: 
+- :commit: 0ed41ef78a807f5d195b070bce7c433fcf6f5cb3
```
mesternefeld: This is not the vcc but rather part of the fix. I'd keep digging back a few commits

cves/CVE-2014-1732.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
```
mesternefeld: I found this vulnerability interesting with how it affects older machines and was fixed by a simple observer pattern. 5 upvotes from me! 🎉

cves/CVE-2013-0891.yml
blackpan2: Looking at this [diff](https://src.chromium.org/viewvc/chrome/trunk/src/webkit/blob/blob_url_request_job.cc?r1=163245&r2=179901&pathrev=179901), this seems to be the VCC.

SVN commit 163245

cves/CVE-2013-0891.yml
```diff
@@ -149,8 +162,10 @@ lessons:
     If you think of another lesson we covered in class that applies here, feel
     free to give it a small name and add one in the same format as these.
   defense_in_depth:
-    applies: 
-    note: 
+    applies: true
+    note: |
+      The solution to this vulnerability including checking the cast values and limits
```
blackpan2: 👍

cves/CVE-2014-1732.yml
blackpan2: Observer is the best +2

PR #149, opened 2018-03-22
CVE-2010-1503 and CVE-2013-2845

lweintraub13:
Round 1 submission for History Project

### DIFF
--- a/cves/CVE-2010-1503.yml
+++ b/cves/CVE-2010-1503.yml
@@ -80,8 +80,8 @@ unit_tested:
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
   answer: It doesn't appear to have any tests for this vulnerability. 
-  code: 
-  fix: 
+  code: false
+  fix: false
 discovered:
   question: |
     How was this vulnerability discovered?
@@ -201,7 +201,7 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those in the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
     The coding mistake that was made seems to be a simple programming error. The programmer in the
     same commit as the VCC used the function EscapeForHTML multiple times prior to the vulnerable line of code. 
     It seems as though he then neglected that the string argument "key" needed to be HTML encoded as well. 
--- a/cves/CVE-2010-1503.yml
+++ b/cves/CVE-2010-1503.yml
@@ -66,7 +66,7 @@ upvotes_instructions: |
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes: 3
 unit_tested:
   question: |
     Were automated unit tests involved in this vulnerability?
@@ -121,10 +121,9 @@ interesting_commits:
   commits:
   - commit: d4b5711eeb1abed75382d9191f603850ac225f28
     note: |
-      A programmer, other than the one who wrote the vulnerability, added
-      a couple lines of code to the same file about 20 lines above the vulnerable, lack
-      of HTML encoding. This commit took place in December 2009, just 4 months
-      before the fix commit. 
+      A programmer, other than the one who wrote the lack of HTML encoding vulnerability, added
+      multiple lines of code to the same file, about 20 lines above the vulnerability. 
+      This commit took place in December 2009, just 4 months before the fix commit. 
 major_events:
   question: |
     Please record any major events you found in the history of this
--- a/cves/CVE-2010-1503.yml
+++ b/cves/CVE-2010-1503.yml
@@ -4,7 +4,7 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: CWE-79
 curated_instructions: |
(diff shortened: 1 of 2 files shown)


cves/CVE-2010-1503.yml
```diff
@@ -114,10 +119,12 @@ interesting_commits:
...
     emerging themes?
   commits:
-  - commit: 
-    note: 
-  - commit: 
-    note: 
+  - commit: d4b5711eeb1abed75382d9191f603850ac225f28
+    note: |
+      A programmer, other than the one who wrote the vulnerability, added
+      a couple lines of code to the same file about 20 lines above the vulnerable, lack
```
lrv1668: Do you mean vulnerability?

cves/CVE-2010-1503.yml
```diff
@@ -190,6 +200,9 @@ mistakes:
...
 
     Use those questions to inspire your answer. Don't feel obligated to answer
-    every one. Write a thoughtful entry here that those ing the software
+    every one. Write a thoughtful entry here that those in the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    The coding mistake that was made seems to be a simple programming error. The programmer in the
+    same commit as the VCC used the function EscapeForHTML multiple times prior to the vulnerable line of code. 
+    It seems as though he then neglected that the string argument "key" needed to be HTML encoded as well.
```
lrv1668: What I find interesting about this commit is that the same programmer sanitized the key in the same file, but just seemed to forget to sanitize the `data->assign` line, it's nice to know that other people can make the same mistakes as us. 

CWE is relevant, and writing and spelling is good except for the comment I added above. I'll give 3 upvotes for this CVE.

cves/CVE-2013-2845.yml
```diff
@@ -193,4 +234,9 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    The developers did not know that that the buffer was being used after it was freed. 
+    I did not see any checks in place to ensure that the buffer was safe to use before 
+    the call. It looks like it was a design mistake and design decision to change OwnPtr to 
+    RefPtr, and recreate the corresponding functions required.
+
```
lrv1668: It's interesting how widespread the Ptr methods were used in this subsystem, makes you think if it would be easier to implement all of these Ptr calls in one file to make it more cohesive, or just to overwrite the Ptr methods in the first place. This must have been a very tedious fix to do! 

CWE is relevant, and grammar and spelling are good. I'll give 3 upvotes for this CVE as well.

cves/CVE-2010-1503.yml
lweintraub13: Yes I do, good catch, I will fix that right now.

cves/CVE-2010-1503.yml
lweintraub13: I agree, just forgetting to sanitize one input can cause a vulnerability, even if all of the other inputs are sanitized. I think a good fix for input sanitization problems like this one is to have more testing to ensure that the input truly is being checked.

cves/CVE-2013-2845.yml
lweintraub13: A very tedious fix to do, and a very tedious hunt I went on to find all of the VCCs. 

I'm not sure how you would implement all of the Ptr calls in one file, so if you can elaborate, I would be interested in listening. I agree with you in thinking that it was interesting that the developer went back in to all of the files to change OwnPtr and Ptr to RefPtr and Ref, instead of just changing the content of the functions so that they didn't have to go back into all of the files that called those functions.

cves/CVE-2010-1503.yml
```diff
@@ -92,10 +97,10 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
-  answer: 
-  date: 
-  automated: 
-  google: 
+  answer: Robert Swiecki from the Google Security Team discovered this vulnerability by trying a simple alert script in the net-internals url.
```
rl5849: Small detail: consider using the '|' so that this sentence can be broken up to match the line length of the the rest of the file.

cves/CVE-2013-2845.yml
```diff
@@ -32,7 +32,10 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  The Web Audio implementation in Google Chrome allows remote attackers to cause
```
rl5849: The first sentence is sort of confusing. I would start with something like, "The implementation of web audio (the aspect of chrome that allows it to play audio over the web)" This offers some more context about chrome. Also 'denial of service due to ...' could be more well said 'denial of service by means of corrupting memory'. Finally 'unknown attack vectors', does that mean the chrome teams knows and wont tell us or that they think there might be another way, but no one really ever tested it.

cves/CVE-2010-1503.yml
lweintraub13: Sounds like an easy fix to make, will do, thanks.

cves/CVE-2013-2845.yml
lweintraub13: I just worked on rewording the description section, with the guidance that prof. meneely discussed in class. thank you!

cves/CVE-2010-1503.yml
rl5849: This vulnerability is quite simple, and I like that you kept your descriptions and notes simple and didn't convolute them with unecessary detail. This CVE is very intersting and an easy mistake for anyone to make, I will give it 3 upvotes.

lweintraub13: Thank you! I updated the upvotes counter

PR #150, opened 2018-03-22
Finished write-up for CVE-2014-1733 and CVE-2015-6762

BoscoJared:
Full write-up for CVE-2014-1733 Vulnerability in Chromium

### DIFF
--- a/cves/CVE-2014-1733.yml
+++ b/cves/CVE-2014-1733.yml
@@ -199,11 +199,4 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: The mistakes made was an overlook in how the merging behavior could
-  be exploited.  The simple overlook allowed the possibility of escaping the
-  sandbox as well as perhaps introducing cyclic behavior during compilation
-  which would cause a crash.  After the issue was brought to attention, the fix
-  was implemented quite cleanly, with unittests to back up the functionality.
-
-  There should be no more issues with that particular piece of merge code as it
-  seemed to be thoroughly fixed.
+  answer: The mistakes made was an overlook in how the merging behavior could be exploited.  The simple overlook allowed the possibility of escaping the sandbox as well as perhaps introducing cyclic behavior during compilation which would cause a crash.  After the issue was brought to attention, the fix was implemented quite cleanly, with unittests to back up the functionality. There should be no more issues with that particular piece of merge code as it seemed to be thoroughly fixed.
--- a/cves/CVE-2014-1733.yml
+++ b/cves/CVE-2014-1733.yml
@@ -99,8 +99,7 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
-  answer: Jed Davis found the compiler bug while integrating the seccomp-bpf
-  codegen into another unrelated project.
+  answer: Jed Davis found the compiler bug while integrating the seccomp-bpf codegen into another unrelated project.
   date: 2014-03-10
   automated: false
   google: false
@@ -112,8 +111,7 @@ subsystem:
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged.
-  answer: The bug was located in the compiler codegen subsystem of seccomp-bpf.
-  The incident was also located in the sandboxing portion of Chromium
+  answer: The bug was located in the compiler codegen subsystem of seccomp-bpf. The incident was also located in the sandboxing portion of Chromium
   name: seccomp-bpf
 interesting_commits:
   question: |
@@ -135,9 +133,7 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: No major events during this time.  It was an area thought to be free
-  of issues, which is why the vulnerability grabbed attention on the Chromium
-  forum.
+  answer: No major events during this time.  It was an area thought to be free of issues, which is why the vulnerability grabbed attention on the Chromium forum.
   events:
   - name:
     date:
@@ -172,9 +168,7 @@ lessons:
     note:
   distrust_input:
     applies: true
-    note: This vulnerability involved merging two instruction blocks, and
-    instruction blocks carefully crafted could force unwanted system calls to
-    be executed outside of the sandbox.
+    note: This vulnerability involved merging two instruction blocks, and instruction blocks carefully crafted could force unwanted system calls to be executed outside of the sandbox.
   security_by_obscurity:
     applies: false
     note:
--- a/cves/CVE-2014-1733.yml
+++ b/cves/CVE-2014-1733.yml
@@ -81,9 +81,7 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: The bug fix was ran against automated tests, and new tests had to be
-  written to showcase the exploit as it was not detected with the unit tests
-  already present
+  answer: The bug fix was ran against automated tests, and new tests had to be written to showcase the exploit as it was not detected with the unit tests already present
   code: true
   fix: true
 discovered:
--- a/cves/CVE-2014-1733.yml
+++ b/cves/CVE-2014-1733.yml
@@ -4,7 +4,7 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE:
+CWE: CWE-94
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
@@ -60,7 +60,7 @@ fixes_vcc_instructions: |
 fixes:
 - :commit: f6205c925db9508cb21c98cf7488eb785e08ff0b
   :note: ''
-vccs: []
+vccs: 46b170d06f1e3d7333d204c0b20a7e8a06259fa4
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
@@ -214,4 +214,4 @@ mistakes:
   was implemented quite cleanly, with unittests to back up the functionality.
 
   There should be no more issues with that particular piece of merge code as it
-  seemed to be thoroughly fixed. 
+  seemed to be thoroughly fixed.
--- a/cves/CVE-2014-1733.yml
+++ b/cves/CVE-2014-1733.yml
@@ -60,7 +60,9 @@ fixes_vcc_instructions: |
 fixes:
 - :commit: f6205c925db9508cb21c98cf7488eb785e08ff0b
   :note: ''
-vccs: 46b170d06f1e3d7333d204c0b20a7e8a06259fa4
+vccs:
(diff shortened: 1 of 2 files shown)


cves/CVE-2014-1733.yml
```diff
@@ -32,7 +32,13 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  A function PointerCompare in the compiler did not properly handle an edge-case
+  when merging two blocks, which means that a remote attacker could potentially
```
andymeneely: Which compiler? Is this Javascript? I could use a bit more details around the scenario.

cves/CVE-2014-1733.yml
```diff
@@ -32,7 +32,13 @@ description_instructions: |
...
+  circumvent the sandbox and use non-secure or non-allowed functionality.
```
andymeneely: What does it mean to violate the sandbox? A specific hypothetical flow of events would be helpful here.

cves/CVE-2014-1733.yml
```diff
@@ -32,7 +32,13 @@ description_instructions: |
...
+
+  The fix was to not merge two basic blocks unless the extensions of each block
```
andymeneely: I didn't understand this sentence. Remember that your reader isn't looking at the fix, so you'll need to do more translation.

cves/CVE-2014-1733.yml
```diff
@@ -47,22 +53,24 @@ reviews:
...
+repo:
 fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: f6205c925db9508cb21c98cf7488eb785e08ff0b
   :note: ''
-vccs: []
+vccs:
+- :commit: 46b170d06f1e3d7333d204c0b20a7e8a06259fa4
```
andymeneely: This doesn't show up in the the Chromium source tree. Is this commit correct?

cves/CVE-2014-1733.yml
```diff
@@ -32,7 +32,13 @@ description_instructions: |
...
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  A function PointerCompare in the compiler did not properly handle an edge-case
+  when merging two blocks, which means that a remote attacker could potentially
+  circumvent the sandbox and use non-secure or non-allowed functionality.
+
+  The fix was to not merge two basic blocks unless the extensions of each block
+  is equivalent.
```
andymeneely: I also saw stuff about `seccomp-bpf codegen` - here would be a good spot to explain what code generation is and how it applies to this vulnerability.

cves/CVE-2014-1733.yml
```diff
@@ -115,10 +123,10 @@ interesting_commits:
     interesting in light of the lessons learned from this vulnerability. Any
     emerging themes?
```
andymeneely: I need to know that you actually tried to do this part. There has to be a reason that there were no interesting commits.

cves/CVE-2014-1733.yml
```diff
@@ -127,12 +135,12 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer: No major events during this time.  It was an area thought to be free of issues, which is why the vulnerability grabbed attention on the Chromium forum.
```
andymeneely: How do you know the area was thought to be free of issues? Is there something to cite for this?

cves/CVE-2014-1733.yml
```diff
@@ -193,4 +201,4 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: The mistakes made was an overlook in how the merging behavior could be exploited.  The simple overlook allowed the possibility of escaping the sandbox as well as perhaps introducing cyclic behavior during compilation which would cause a crash.  After the issue was brought to attention, the fix was implemented quite cleanly, with unittests to back up the functionality. There should be no more issues with that particular piece of merge code as it seemed to be thoroughly fixed.
```
andymeneely: So it's unclear to me what the "merging behavior" is here. I could use more explanation. Also, what I need in this paragraph is an understanding of how this problem was MISSED. What happened between the VCC and the fix? How did people miss it? Were people working on similar issues or not?

cves/CVE-2015-6762.yml
```diff
@@ -32,7 +32,15 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: When fetching a font from a remote resource, the application would
+  always utilize CORS; however, making a same-origin request would not use CORS
```
andymeneely: Your audience won't know what CORS is, nor same-origin. Both could use some explanation.

cves/CVE-2015-6762.yml
```diff
@@ -32,7 +32,15 @@ description_instructions: |
...
+  thus allowing the font url to be bypassed from another remote server if the
+  same-origin forwarded the request.  This allowed things like forwarding to
+  another server to grab a font and apply it to the DOM, where the application
```
andymeneely: DOM needs explanation

cves/CVE-2015-6762.yml
```diff
@@ -48,22 +56,24 @@ reviews:
...
+repo:
 fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: 212b92858f0ae007c90abd10b40ba805a13af43c
   :note: ''
-vccs: []
+vccs:
+- :commit: 8add030475be094a47b0ec2abf7bfcf34bb8079f
```
andymeneely: If there was never an introduction, then what's this commit?

cves/CVE-2015-6762.yml
```diff
@@ -48,22 +56,24 @@ reviews:
...
+  :note: There was never an introduction of the vulnerability as they never used CORS on same-origin requests, so the commit points to when the function was first added to the file ( fetch function )
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
   For the second round of reviewing, you will be giving a certain amount of
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes:
```
andymeneely: +4 from me.

cves/CVE-2015-6762.yml
```diff
@@ -194,4 +204,4 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer:
```
andymeneely: As before, make sure the mistake cover the process in addition to the design. I would also say that there's a design choice, not just a coding choice here.

cves/CVE-2015-6762.yml
jedjas12: +2. The idea behind this exploit is very smart to me, if I understand it correctly. They used resources that were considered inside of the trust boundary so they wouldn't be checked, then had it call out to a remote server so they could implement the attack?

cves/CVE-2015-6762.yml
```diff
@@ -116,10 +126,10 @@ interesting_commits:
     interesting in light of the lessons learned from this vulnerability. Any
     emerging themes?
   commits:
-  - commit: 
-    note: 
-  - commit: 
-    note: 
+  - commit:
```
jedjas12: Did they have no interesting commits? Or were there no commits for this between them implementing and then fixing the bug?

cves/CVE-2015-6762.yml
```diff
@@ -194,4 +204,4 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: The mistake was in the assumption that a same-origin request would stay on the same origin and therefore CORS would not be needed.  It is safer to just apply CORS to any request due to the fact that there could be redirects.  After being made aware of the issue, it was a simple fix, and there are not many exploits possible with such a vulnerability - still, the lesson of not trusting input holds true.
```
jedjas12: Formatting comment: This would be easier to read with a pipe (|) on multiple lines, rather than one long side-scroll.

Question: Did they ever state why they didn't do the CORS check on the same-origin request? Was it a time/complexity issue? Oversight? Hubris?

cves/CVE-2015-6762.yml
rsauberman: +4 from me. Very interesting, especially because it was a simple fix to apply CORS to all requests.

rsauberman: +4 from me. Very interesting and it was a simple fix to apply CORS to all requests.

cves/CVE-2014-1733.yml
```diff
@@ -47,22 +53,24 @@ reviews:
...
+  :note: ''
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
   For the second round of reviewing, you will be giving a certain amount of
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes:
```
rsauberman: +4 from me. Very interesting and it was a simple fix to apply CORS to all requests.

cves/CVE-2015-6762.yml
rsauberman: +4 from me. Very interesting and it was a simple fix to apply CORS to all requests.

cves/CVE-2015-6762.yml
BoscoJared: Yes, in a project like Chromium the choice to not apply CORS on same-origin was definitely a conscious design decision that ended up causing a vulnerability.  I've updated it to reflect that.

cves/CVE-2015-6762.yml
BoscoJared: I added an explanation for CORS, and same-origin in the description

cves/CVE-2015-6762.yml
BoscoJared: I added an explanation for DOM in the description

cves/CVE-2015-6762.yml
BoscoJared: added

BoscoJared: Yes, the designers assumed that a same-origin request could leave out the CORS headers, and thus when redirected within that resource fetch, it could go anywhere and be exploited from that angle.  

Interestingly enough, in the discussion they talked about how Safari doesn't apply CORS to any of its font fetches ( but maybe they do now ) which made them not really think it was a crazy vulnerability, but obviously they did end up patching it.

cves/CVE-2015-6762.yml
BoscoJared: I fixed the same line issue with the pipe, thanks.

It wasn't stated exactly why they made this assumption, but it did appear to be a conscious design decision that was exploited, and not an oversight or anything like that.

cves/CVE-2015-6762.yml
BoscoJared: Thanks for the feedback, added the upvotes

PR #151, opened 2018-03-22
 CVE-2010-3251 and CVE-2016-1632

peckrog:


hnv1002:
CVE-2010-3251
- The description for this vulnerability is very specific to Chrome. You can explain what WebSocket does and then what causes Chrome to crash specifically. Looks like the WebSocket deferences to a null pointer during a host resolution (reloading a page for example), the code forgets to check for the state of socket stream before opening a new connection, thus continuing to use a null object.
- From my findings, this vulnerability was found based on a reproduction of another bug, perhaps they were running regression tests and found this bug. https://bugs.chromium.org/p/chromium/issues/detail?id=37190
- I think defense in depth can be applied here. All objects need to be checked before used to avoid referencing a null pointer. Also, they should have catched and handled the exception properly instead of letting it crash the application.
- upvotes: 2

CVE-2016-1632
- The description is very vague. How does bypassing intended access restrictions cause incorrect cast? It looks like they prevent normal users from seeing the bug report, perhaps you can explain a little bit more or come up with some scenarios in which this can happen.
- Since this vulnerability has to do with access restrictions, I think least_privilege principle can be applied here. 
- upvotes: 4

cves/CVE-2010-3251.yml
```diff
@@ -32,7 +32,10 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+    The WebSockets implementation in Google Chrome allowed attackers to
```
moseslagoon: Would be great, if you could have a more general description of the vulnerability instead of the specifics. What the websocket does? Any particular scenario this vulnerability would be used by an attacker? If you could touch on an example? I think that would be more helpful.

cves/CVE-2010-3251.yml
```diff
@@ -92,10 +98,13 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
-  answer: 
-  date: 
-  automated: 
-  google: 
+  answer: |
```
moseslagoon: As Chris mentioned, seems that the vulnerability was found based on a reproduction of another bug.

cves/CVE-2010-3251.yml
```diff
@@ -52,8 +55,10 @@ fixes_vcc_instructions: |
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: af60ef0bb32bd0babdf51e0e11795ad61b30ec4d
-  :note: ''
-vccs: []
+  :note:
+vccs: 
+- :commit: aef03130359ef824153b4ed088622460bbccdb0f
+  :note:
 upvotes_instructions: |
```
moseslagoon: Upvote: 2

cves/CVE-2010-3251.yml
```diff
@@ -192,4 +201,10 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
```
moseslagoon: Along with the lack of unit tests,  I thought that this would be a good design problem as well. I like how you touch on the lack of automated testing? I think a good design of the system would have led to the coverage of these cases. I might be wrong but something to think about?

cves/CVE-2010-3251.yml
moseslagoon: this might help: https://bugs.chromium.org/p/chromium/issues/detail?id=51846

cves/CVE-2016-1632.yml
```diff
@@ -32,7 +32,9 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  Attackers can bypass intended access restrictions using JavaScript code.The attack
```
moseslagoon: A little bit of a longer example and elaboration would help. The description seems very high level. I think example on how bypassing would work in this case. Perhaps, a scenario would be much more helpful in my opinion. Seems vague.

cves/CVE-2010-3251.yml
moseslagoon: I was having a permission issues accessing the bug report https://bugs.chromium.org/p/chromium/issues/detail?id=549986. Usually, the report is lot more helpful in terms of summarizing and rephrasing the issue as I got most of my understanding from it. What that an issue for you as well?

cves/CVE-2016-1632.yml
moseslagoon: I was having permission issue accessing the bug report https://bugs.chromium.org/p/chromium/issues/detail?id=549986. Usually, the report is lot more helpful in terms of summarizing and rephrasing the issue as I got most of my understanding from it. What that an issue for you as well?

cves/CVE-2016-1632.yml
```diff
@@ -74,9 +80,11 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
```
moseslagoon: This seems to be an interesting issue. I would give it an upvote : 4

cves/CVE-2016-1632.yml
```diff
@@ -92,7 +100,8 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
-  answer: 
+  answer: |
+    Unfortunately I don't have access to the issue page for this so I do not know.
```
moseslagoon: Ignore my previous question on being able to access the page. But for the description https://nvd.nist.gov/ also had a pretty good outline of what the vulnerability is.

PR #158, opened 2018-03-22
CVE-2016-5130 and CVE-2016-1679

BenjaminHunt:
CVE-2016-5130 complete
CVE-2016-1679 complete

### DIFF
--- a/cves/CVE-2016-1679.yml
+++ b/cves/CVE-2016-1679.yml
@@ -118,8 +118,8 @@ subsystem:
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged.
-  answer: content/child/
-  name: 
+  answer:
+  name: content/child/
 interesting_commits:
   question: |
     Are there any interesting commits between your VCC(s) and fix(es)?
--- a/cves/CVE-2016-1679.yml
+++ b/cves/CVE-2016-1679.yml
@@ -130,8 +130,11 @@ interesting_commits:
     interesting in light of the lessons learned from this vulnerability. Any
     emerging themes?
   commits:
-  - commit: 
-    note: 
+  - commit: b5bdf3778209179111c9f865af00940e74aa20e7
+    note: |
+    I thought it was interesting how the solution was to remove the setter and
+    change the logging method. It reminded me of "setters and getters are evil"
+    but was surprised that it was able to cause a denial of service.
   - commit: 
     note: 
 major_events:
@@ -208,4 +211,7 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: nil
+  answer: |
+    A design/coding mistake has been made for this section of code. A setter
+    was introduced in an area where there was no need or reason for it, which
+    lead to a denial of service by crashing the site.
--- a/cves/CVE-2016-1679.yml
+++ b/cves/CVE-2016-1679.yml
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: Other (NVD-CWE-Other)
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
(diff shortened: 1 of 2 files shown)


cves/CVE-2016-5130.yml
```diff
@@ -60,7 +74,9 @@ fixes:
   :note: ''
 - :commit: 4d78d33cb1852e7d487f1d327517c4c4044a3ca3
   :note: ''
-vccs: []
+vccs:
```
meyersbs: The VCC(s) are very important for curating this data -- try and do a little digging and pinpoint when the Blink dependency was introduced.

cves/CVE-2016-5130.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: Improper Access Control
```
meyersbs: Please include the CWE number here as well.

BenjaminHunt: Done.

cves/CVE-2016-5130.yml
BenjaminHunt: I've briefly looked into it a little bit, but it is difficult to find. There are 15,200 code results, and 209,967 commit results when I search "Blink." The VCE points to the file that addresses the problem, not the file that uses it or calls the dependency, so I don't know how to find it.

If I think of a way to find where the dependency was introduced I will do it. If you have any suggestions, please let me know.

cves/CVE-2016-1679.yml
```diff
@@ -32,7 +32,10 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  The ToV8Value function has unneccessary calls of setter functions that may
+  crash the Google Chrome Web Browser. This is do to a lack of necessary
```
Josh-Akers: Incorrect use of the word "do", should be "due".

cves/CVE-2016-1679.yml
```diff
@@ -95,20 +100,23 @@ discovered:
...
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: |
+    A class used V8ValueConverter to convert a DictionaryValue to a v8::Value
+    with an untrustworth ScriptContext. This allows attackers to use a setter
```
Josh-Akers: Spelling error, add a 'y' to the end of "untrustworth".

cves/CVE-2016-1679.yml
```diff
@@ -32,7 +32,10 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
```
Josh-Akers: Description feels like it's missing something.  What does this vulnerability allow and why is it a problem?

The initial bug report here should be of some help.  

https://bugs.chromium.org/p/chromium/issues/detail?id=606390

cves/CVE-2016-1679.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: Other (NVD-CWE-Other)
```
Josh-Akers: I was under the impression that the CWE was supposed to be a number categorized by cwe.mitre.org.

Just go through the list for something that's related to memory not being freed.  

http://cwe.mitre.org/data/definitions/699.html

cves/CVE-2016-5130.yml
Josh-Akers: Not sure if this helps but in 09/2015 the Blink repository was merged into Chromium.

cves/CVE-2016-5130.yml
```diff
@@ -99,10 +111,11 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
-  answer: 
-  date: 
-  automated: 
-  google: 
+  answer: |
+    A chromium developer discovered the flaw in the
+  date: 2013-04-30
```
Josh-Akers: Are you sure this is accurate?  That's three years before the bug was reported.  Also, if this bug is being caused by the added dependency of Blink to chromium, then this bug wouldn't have existed prior to the introduction of those new dependencies would it?

cves/CVE-2016-5130.yml
```diff
@@ -60,7 +70,9 @@ fixes:
   :note: ''
 - :commit: 4d78d33cb1852e7d487f1d327517c4c4044a3ca3
   :note: ''
-vccs: []
+vccs:
+- :commit:
+  :note: 'Whenever the Blink dependency was introduced'
 upvotes_instructions: |
```
Josh-Akers: Upvotes: 4

cves/CVE-2016-5130.yml
BenjaminHunt: That might help, I'll look into it. Thanks!

cves/CVE-2016-5130.yml
BenjaminHunt: I will investigate this further.

cves/CVE-2016-1679.yml
BenjaminHunt: It appears that I was looking in the wrong section. That's what it said under the Technical Details, the vulnerability type.

Under the Evaluator Description, it said CWE-416: Use After Free. I will correct this.

cves/CVE-2016-5130.yml
BenjaminHunt: I believe I read the wrong bug report. Good find.

PR #161, opened 2018-03-23
CVE-2011-3961 and CVE-2016-5171

bxsmith007:
Initial pull request. The edited yaml files are CVE-2011-3961 and CVE-2016-5171.


cves/CVE-2016-5171.yml
```diff
@@ -76,9 +84,12 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
+    From what it looks like, there are two tests that are being used to test if
+    this code is now working. I couldn't find if these tests were there created
```
Michael-Longley: This sentence is a bit confusing can you reword the last part about when the test were created.

cves/CVE-2011-3961.yml
```diff
@@ -32,7 +32,9 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  Race condition in Google Chrome allows remote attackers to execute arbitrary
+  code via vectors that trigger a crash of a utility process.
```
Michael-Longley: Can you say which utility process without too much jargon?

cves/CVE-2011-3961.yml
```diff
@@ -115,10 +129,25 @@ interesting_commits:
...
+      What seems interesting about this commit is there is a just a large amount
+      of code added all at once to to this file. Though it may be correct code,
+      it still seems to have the potential something wrong with it without being
+      thoroughly being tested.
+  - commit: 44e1fbf0fd0a14a375ebbbc56392956683f6261c
+    note: |
+      What is interesting here is that the commit messages talkes about moving
+      the IndexedDB code to its own directory inside of content/... The race
+      condition occurred because of an error with IndexedDB, so this could have
+      been the initial cause of the error.
```
Michael-Longley: You mention here that this commit could be the initial cause of the error however it was not listed in your VCC, can you justify this?

cves/CVE-2011-3961.yml
```diff
@@ -115,10 +129,25 @@ interesting_commits:
...
+  - commit: fc7b441f6e9541d111fadd050d0cbdb6ac761614
+    note: |
+      This commit deals with cleaning up the in_process_webkit, which contained
+      this vulnerability. This clean up could have potentially caused the
+      vulnerability to occur when it could have been correct before.
```
Michael-Longley: Similarly you say this could have caused the vulnerability to appear without listing it in your VCC list, can you explain or add it.

cves/CVE-2016-5171.yml
bxsmith007: What I meant to say was: "I couldn't find if these tests were created before or after the finding of the vulnerability."

cves/CVE-2011-3961.yml
```diff
@@ -93,20 +103,24 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave this part blank.
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: |
+    It was not really clear how the vulnerability was initally found. It seems
```
dwg7486: It's probably safe to say they're a bug hunter - normal users of Chromium probably wouldn't report such a nuanced bug, they'd just switch browsers.  So this person was looking to break Chromium and that's how they discovered this.

cves/CVE-2016-5171.yml
```diff
@@ -194,4 +222,14 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    What seemed to lead to this vulnerability was using there own testing
```
dwg7486: Wrong use of `there`.

cves/CVE-2011-3961.yml
bxsmith007: The only reference of this that I see is in the main commit when it says, "A few clients of the utility process..." After a little google searching, google has a general utility process, so I will change my description to be "the utility process" rather than "a utility process".

cves/CVE-2011-3961.yml
bxsmith007: I believe I did mention this commit in my list of VCC's.

cves/CVE-2011-3961.yml
bxsmith007: Ya, I think you are right. Looking at the chat for bug fixes, it did seem to know about the bounty and how to go about finding the fix in is own time. This would then make sense that this person was a bug hunter. Thank you for the input!

cves/CVE-2016-5171.yml
bxsmith007: Thank you! I will fix that.

cves/CVE-2011-3961.yml
bxsmith007: Looking at this more carefully, I don't think this is the cause. I thought it was interesting that there was a movement of the IndexedDB to its own directory, but this commit had nothing to do with the problem code. I will probably delete this commit. Thank you!

cves/CVE-2011-3961.yml
```diff
@@ -47,22 +49,28 @@ reviews:
...
+  :note: ''
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
   For the second round of reviewing, you will be giving a certain amount of
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes:
```
BrendanJones44: +3 upvotes from me!

cves/CVE-2016-5171.yml
```diff
@@ -48,22 +52,26 @@ reviews:
...
+- :commit: ebff986b4bd9cbe0b0328adaf86dc69964945173
+  :note: ''
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
   For the second round of reviewing, you will be giving a certain amount of
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes:
```
BrendanJones44: +1 upvote from me!

PR #164, opened 2018-03-23
CVE-2011-3898 and CVE-2015-1257

JoncarloLovesSecurity:
Initial draft and PR of the YAML files for CVE-2011-3898 and CVE-2015-1257

### DIFF
--- a/cves/CVE-2011-3898.yml
+++ b/cves/CVE-2011-3898.yml
@@ -22,11 +22,9 @@ announced: 2011-11-10 06:55:02.787000000 -05:00
 description_instructions: |
   You can get an initial description from the CVE entry on cve.mitre.org. These
   descriptions are a fine start, but they can be kind of jargony.
-
   Rewrite this description in your own words. Make it interesting and easy to
   read to anyone with some programming experience. We can always pull up the NVD
   description later to get more technical.
-
   Try to still be specific in your description, but remove Chromium-specific
   stuff. Remove references to versions, specific filenames, and other jargon
   that outsiders to Chromium would not understand. Technology like "regular
@@ -55,7 +53,7 @@ fixes_vcc_instructions: |
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: eb61fc8d78c9083d33bb3d214a98da81a22948ab
-  :note: ''
+  :note: 'Removed the upper-bound on the Java version check'
 vccs: [7f18717e883b61612ef851695d173021aeb65cf5]
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
@@ -64,7 +62,7 @@ upvotes_instructions: |
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes: 2
 unit_tested:
   question: |
     Were automated unit tests involved in this vulnerability?
@@ -120,7 +118,7 @@ interesting_commits:
     Write a brief (under 100 words) description of why you think this commit was
     interesting in light of the lessons learned from this vulnerability. Any
     emerging themes?
-  commits: 89ec0bc44ba3fec94e4c6d02366cfef72a5c8eb4
+  commits:
   - commit: 89ec0bc44ba3fec94e4c6d02366cfef72a5c8eb4
     note: This commit is between the VCC and the commit that fixed this vulnerability.
           It demonstrates that this code was maintained pretty regularly but developers
--- a/cves/CVE-2011-3898.yml
+++ b/cves/CVE-2011-3898.yml
@@ -32,9 +32,9 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: If a user had a certain version of Java applets were allowed 
-             to run without user confirmation. This was due to a misconfiguration
-             error caused by the choice to only check Java versions below 7.
+description: If users had a version of Java that was 7 or above, then applets 
+             were allowed to run without any confirmation from the user. 
+             This was due to the code only checking for Java versions below 7.
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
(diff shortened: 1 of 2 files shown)


cves/CVE-2011-3898.yml
```diff
@@ -54,7 +56,7 @@ fixes_vcc_instructions: |
 fixes:
 - :commit: eb61fc8d78c9083d33bb3d214a98da81a22948ab
   :note: ''
-vccs: []
+vccs: [89ec0bc44ba3fec94e4c6d02366cfef72a5c8eb4]
```
JohnPMurray: This vcc looks like they just updated the versions. The whole version checking system looks like it was implemented in commit 7f18717e883b61612ef851695d173021aeb65cf5. That one might be more of a vcc than this one is. Maybe this one could go in the "interesting commits" section

cves/CVE-2011-3898.yml
```diff
@@ -32,7 +32,9 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: If a user had a certain version of Java applets were allowed 
+             to run without user confirmation. This was due to a misconfiguration
+             error caused by the choice to only check Java versions below 7.
```
JohnPMurray: Wording here was a little confusing to me. Maybe break up the second sentence into 2 sentences (ie. This was due to a misconfiguration error. The system was only checking Java versions below 7.)

cves/CVE-2011-3898.yml
```diff
@@ -32,7 +32,9 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: If a user had a certain version of Java applets were allowed
```
JohnPMurray: Add a comma after "Java"

cves/CVE-2015-1257.yml
```diff
@@ -96,20 +100,24 @@ discovered:
...
-  automated: 
-  google: 
-  contest: 
+  answer: The  crash was discovered through the use of a fuzzer designed by
+          bug bounty hunter miaubiz. The fuzzer was designed to target the
+          SVG component the Blink engine.
+  date: 2015-03-18
+  automated: true 
+  google: false
+  contest: miaubiz
```
JohnPMurray: I think this is supposed to be the name of a contest, not a person. It doesn't look like the bug was found in a contest, just an individual on their own time.

cves/CVE-2011-3898.yml
JoncarloLovesSecurity: After reviewing that commit, I definitely agree it deserves to be the main vcc. I will consider keeping 89ec0bc44ba3fec94e4c6d02366cfef72a5c8eb4 as an interesting commit and perhaps comment on how easy it was to overlook this mistake.

cves/CVE-2011-3898.yml
JoncarloLovesSecurity: I agree that this description sounds clunky and confusing. I think I will change it to "If users had a version of Java that was 7 or above, then applets were allowed to run without any confirmation from the user. This was due to the code only checking for Java versions below 7". Hopefully this removes any confusion.

cves/CVE-2011-3898.yml
JoncarloLovesSecurity: I addressed this error in other comments.

cves/CVE-2015-1257.yml
JoncarloLovesSecurity: I believe you are right. I will remove miaubiz from the contest field and make sure he is mentioned elsewhere in the CVE report.

cves/CVE-2011-3898.yml
```diff
@@ -127,7 +136,9 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer: When this vulnerability was introduced many assumptions were made about
+          what kind of input the code in the SVG module would recieve. This lead
```
KylerF: The last sentence doesn't read quite right. Needs a grammar fix at the end.

cves/CVE-2011-3898.yml
```diff
@@ -127,7 +136,9 @@ major_events:
...
+          to a bug be introduced the code very early on.
```
KylerF: Perhaps the case for defense-in-depth could be made as well. The developers seemed to include an upper bound out of convenience, ignoring the fact that Java would surpass that upper bound. It would follow defense-in-depth to instead recognize this possibility and defend against it.

cves/CVE-2011-3898.yml
```diff
@@ -176,8 +187,11 @@ lessons:
     applies: 
     note: 
   yagni:
-    applies: 
-    note: 
+    applies: true
+    note: This bug was introduced by the fact that a developer
```
mikex5: very interesting to see this kind of vulnerability! +2 upvotes

cves/CVE-2011-3898.yml
JoncarloLovesSecurity: I edited the last sentence so that it made sense grammatically. I also added more to the answer to provide more information.

JoncarloLovesSecurity: The answer was actually somehow misplaced and belonged in CVE-2015-1257. Not sure how I made that mistake but I corrected it.

cves/CVE-2015-1257.yml
```diff
@@ -49,15 +53,17 @@ reviews:
...
-fixes_vcc_instructions: |
-  Please put the commit hash in "commit" below (see my example in
-  CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: 34d7c7dc4327329632881a6f8ce866d49d756eb3
-  :note: ''
+  :note: 'Committed on April 16, 2015. The initial fix did not
+          work as expected and was reverted. The new fix added
+          checks at all locations the vector is used because
+          it can not be guarenteed that a vector will contained
```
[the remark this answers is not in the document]
JoncarloLovesSecurity: Fixed.

PR #182, opened 2018-09-24
Add nicknames to CVE yml files

andymeneely:
A "nickname" is a plain-English title for a given vulnerability that would be shown alongside its CVE.


cves/CVE-2013-6665.yml
```diff
@@ -1,5 +1,6 @@
 ---
 CVE: CVE-2013-6665
+nickname: The Example
```
andymeneely: Let's actually name this. Let's go for "Texture Smash"

PR #191, opened 2018-10-25
CVE-2015-1240 & CVE-2016-1686

ewd9627:
Investigation of two issues, the first one stemmed from an issue with an out-of-bounds error in the WebGL functions, and the second stemmed from an issue with an out-of-bounds error in the PDFium interpreter.

### DIFF
--- a/cves/CVE-2015-1240.yml
+++ b/cves/CVE-2015-1240.yml
@@ -33,11 +33,11 @@ description_instructions: |
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
 description: |
-  In Google Chrome, prior to the 42.0.2311.90 release, there was a vulnerability
-  which allowed remote attackers to cause a denial of service (out-of-bounds read)
-  via a crafted WebGL program that triggers a state inconsistency. In other
-  words, any attacker writing a WebGL program who is aware of this bug may force
-  an integer out-of-bounds read which may halt or break the WebGL operation.
+  There was a vulnerability which allowed remote attackers to cause a denial of
+  service (out-of-bounds read) via a crafted WebGL program that triggers a state
+  inconsistency. In other words, any attacker writing a WebGL program who is
+  aware of this bug may force an integer out-of-bounds read which may halt or
+  break the WebGL operation.
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
@@ -90,8 +90,9 @@ unit_tested:
     It is unclear if automated unit tests were involved with this vulnerability,
     however, they were certainly involved in the detection and resolving of this
     vulnerability.
-  code:
-  fix:
+  code: An unhandled exception in the OpenGL handler caused this issue
+  fix: |
+  Validating inputs and parameters in order to avoid improper code causing hangs
 discovered:
   question: |
     How was this vulnerability discovered?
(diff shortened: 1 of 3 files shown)


cves/CVE-2015-1240.yml
```diff
@@ -81,9 +86,13 @@ unit_tested:
...
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
+    It is unclear if automated unit tests were involved with this vulnerability,
+    however, they were certainly involved in the detection and resolving of this
+    vulnerability.
+  code: An unhandled exception in the OpenGL handler caused this issue
+  fix: |
```
andymeneely: Looks like this is one of the culprits of your YML syntax not building. Remember that getting CI to pass is the main part of the Part 1 grade.

cves/CVE-2016-1686.yml
```diff
@@ -79,9 +85,39 @@ unit_tested:
...
-  code: 
-  fix: 
+  answer: |
+    PDFium comes with many unit tests, such as pdfium_unittests and
+    pdfium_embeddertests, run_corpus_tests.py, run_pixel_tests.py, and
+    run_javascript_tests.py.
+
+    The fix involved adjusting the unit tests once the bug was noticed, and
+    testing once adjustments were made to ensure the fix was proper.
+  code: |
```
andymeneely: These need to be true or falses. The "answer" part can have code in it, but the code flag needs to be true or false.

cves/CVE-2016-1686.yml
```diff
@@ -79,9 +85,39 @@ unit_tested:
...
+           m_nComponents = static_cast<uint32_t>(comps);
+           if (m_Family == PDFCS_LAB && m_nComponents != 3) {
+            m_pCompData = nullptr;
+             return 0;
+           }
+           m_pCompData = GetDecodeAndMaskArray(m_bDefaultDecode, m_bColorKey);
+           if (!m_pCompData) {
+             return 0;
+           }
+  fix: |
```
andymeneely: same here: this fix field needs to be true or false, and then the answer needs to be readable in prose.

This is more of a writing assignment anyway. We'll present links to the code that people can follow, so explaining what happened is more valuable than copying code snippets without much context.

cves/CVE-2016-1686.yml
```diff
@@ -200,4 +259,8 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    In my opinions, the mistakes were mostly coding mistakes, where ways of
+    avoiding issues such as this were already known to the coding world, however
```
andymeneely: This paragraph could be more specific. Picture yourself in a meeting where you had made this mistake and explain what you'd do to prevent it in the future.

cves/CVE-2016-1686.yml
ewd9627: I realize I have fixed this prior to responding to the comment - thank you for this, I was not 100% sure what was expected here when reading the instructions!

cves/CVE-2016-1686.yml
ewd9627: I realize I have fixed this prior to responding to the comment - thank you for this, I was not 100% sure what was expected here when reading the instructions!

cves/CVE-2015-1240.yml
ewd9627: Yep, the big issue was that I didn't realize that code and fix were supposed to evaluate to true or false or nil, I ended up describing it, and after a few iterations, realized that they should both only have a one word answer. Haha

cves/CVE-2016-1686.yml
ewd9627: I realize I have fixed this prior to responding to the comment - thank you for this, I was not 100% sure what was expected here when reading the instructions!

PR #194, opened 2018-10-25
CVE-2010-2105 and CVE-2015-1278

jahn621:
added changes to CVE-2015-1278 and CVE-2015-1278 setting their curated to true


andymeneely:
Looks like the build didn't run properly the first time. I reran it just now and got these errors: https://travis-ci.org/VulnerabilityHistoryProject/chromium-vulnerabilities/builds/446393209#L503

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
andymeneely: Please delete file this from your pull request

cves/CVE-2010-2105.yml
```diff
@@ -32,15 +32,18 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: Safety specification requirement is not fulfilled within URL Canonicalization.
+Canonicalization is the process of picking the best URL when there are several choices,
```
TylerLima: Is this in terms of handling a search request of a specific site from the home page? If so, I'm wondering (as you probably were to) what sort of impact that can cause to the User or the website runner.

cves/CVE-2010-2105.yml
```diff
@@ -53,8 +56,9 @@ fixes:
 - :commit: 69ee9c92af91fdad5aabf0f5696cd0432102985f
   :note: ''
 vccs:
-- :commit:
-  :note: 
+- :commit: 7753c3f48f5d3fea1ac91575de75d78f8085be83
+  :note: There was a change in logic, so large portions of the file with the vulnerability
```
TylerLima: Was there a sign when they recognized that there logic was flawed or was it a matter of correction to their logic to incorporate Safe Browsing Specialization rules?

cves/CVE-2010-2105.yml
```diff
@@ -93,20 +97,28 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: |
```
TylerLima: Good catch on these details of the vulnerability discovery. Kind of helps answer my earlier question.

cves/CVE-2010-2105.yml
```diff
@@ -152,38 +170,45 @@ lessons:
...
     note: 
   native_wrappers:
-    applies: 
+    applies: false
     note: 
   distrust_input:
-    applies: 
-    note: 
+    applies: true
+    note: the concept of url canonicalization is the idea of distrusting whatever the user is putting
```
TylerLima: This is a good note on perspective of handling the responsibility of the user to handle dangerous inputs as we've discussed in class.

cves/CVE-2015-1278.yml
```diff
@@ -32,7 +32,13 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
```
TylerLima: You've done a great job describing what each of the lesser known details of what is happening very well. This made understanding your CVE's much easier.

cves/CVE-2015-1278.yml
```diff
@@ -97,20 +105,24 @@ discovered:
...
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: |
+    While using an html, authentication php, user response php, and pdf php file in addition to a
+    pdf viewed through chrome, a user was able to create an instance where the pdf gave way to a user's
+    age. Albeit, while the site redirects where it needed to, the overall information that may be falsely
```
TylerLima: This is very strange. I'm curious as to how they went about this based on the pdf php file interacting with the page. If you know anything in particular about this I'd love to know your thoughts on it.

cves/CVE-2015-1278.yml
```diff
@@ -156,37 +168,41 @@ lessons:
...
-    applies: 
+    applies: false
     note: 
   distrust_input:
-    applies: 
-    note: 
+    applies: true
+    note: |
+      removes the ability for the user to interact with the system in ways that do not want it to be touched in.
+      Removes all ability to spoof as other users, potentially ruining the integrity of system access logs.
```
TylerLima: This will likely impact any further CVE's that come up in the future (especially ones that are related to repudiation of the system's integrity).

cves/CVE-2015-1278.yml
```diff
@@ -200,4 +216,14 @@ mistakes:
...
-  answer: 
+  answer: |
+    This vulnerability was most likely created through a design oversight. Largely just because
+    this wasn't an issue found through the Google Chrome developers, but rather a random stranger. It's a
+    very specific way to spoof authentication. Regardless, it ultimately leads to a way for users to input
+    disingenuous data into a system's logs due to being unable to verify where the information comes from.
+
+    The CWE entry most related to this does not have a mitigation. Claiming that the overall broad nature
+    of the vulnerability is more alike a category rather than being a specific fix.
+
```
TylerLima: Good thought as to why the CVE might have existed in the first place, I feel like many of the CVE's follow this pattern (obviously) because they (as devs) need to pick and choose their battles in order to mitigate the greater threats in order of magnitude.

cves/CVE-2010-2105.yml
jahn621: In regards to the search request, while canonicalization may occur specifically to home pages, it wouldn't necessarily be isolated to the home page. However, it should be noted that home pages often have various other links that do lead to the very same home page, and this could potentially lead to repudiation becoming a larger issue if a web crawler is used. Without taking the necessary steps to resolve the vulnerability, attackers could potentially gain access and information they are not meant to have.

cves/CVE-2010-2105.yml
jahn621: It was both actually. There was a large portion of the original thread that was meant to include theory crafting of what was actually wrong and that it merited investigation. On the other end of the spectrum, the other being thing was going over the code and realizing that it didn't fit the same incorporation of the specialization rules.

cves/CVE-2010-2105.yml
jahn621: Never trust the user, not even yourself.

cves/CVE-2015-1278.yml
jahn621: I expanded the original answer, and made another commit to append to the original information that I had. However, the original bug report seems to be missing or  lost to history. The software team would only take the original steps and demonstrate how to recreate the issue within the vulnerability.

cves/CVE-2015-1278.yml
jahn621: Then it's damn good that the vulnerabilities have been fixed then!

cves/CVE-2015-1278.yml
jahn621: I think that there are many other vulnerabilities that the CVEs demonstrate that follow the CWEs far better than this CVE does. Hopefully, in the future the overall classification of this CVE may be split several ways around just to break it down to fit various other CWEs

PR #195, opened 2018-10-25
CVE-2013-0831 and CVE-2015-6781

peterhoward45:


cves/CVE-2013-0831.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: CWE-22 Path Traversal
```
michaelboc: We may have to confirm with the professor; however, I believe that this should only have the number of the CWE.

cves/CVE-2013-0831.yml
```diff
@@ -22,179 +22,185 @@ announced: 2013-01-15 16:55:01.993000000 -05:00
...
   description later to get more technical.
-
   Try to still be specific in your description, but remove Chromium-specific
   stuff. Remove references to versions, specific filenames, and other jargon
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  Directory traversal vulenrability that allows remote attackers to
```
michaelboc: Maybe you should use the term "Path Traversal" for consistency with the well known term for this vulnerability. 

Also, you have miss-spelt "vulnerability".

cves/CVE-2013-0831.yml
```diff
@@ -22,179 +22,185 @@ announced: 2013-01-15 16:55:01.993000000 -05:00
...
-repo: 
+repo:
 fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
-- commit: 
-  note: 
+- commit: 730258b836c753f54901439507fe258babf51e94
+  note: Fix directory traversal in extension_resources.cc.  Adds test case,
```
michaelboc: It would be interesting to read details how the code was patched to mitigate this vulnerability.

cves/CVE-2015-6781.yml
```diff
@@ -47,26 +49,29 @@ reviews:
...
 fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: d02508f8c010b1415f5f1521a6f4cee260d1f8a9
-  :note: ''
+  :note: 'Fix the win64 GN build after r351207.'
 - :commit: 3e7d8e08a0fd550e03c5d83621871bd36803e727
-  :note: ''
+  :note: 'Pull sfntly from GitHub instead of code.google.com. Pick up latest
```
michaelboc: What are "sfntly"?

cves/CVE-2015-6781.yml
```diff
@@ -156,38 +172,43 @@ lessons:
...
+    note:
   yagni:
-    applies: 
-    note: 
+    applies:
+    note:
   complex_inputs:
-    applies: 
-    note: 
+    applies:
```
michaelboc: Could this vulnerability also be a example of complex inputs? Font data appears to be complex with multiple vectors that could be used in an attack.

cves/CVE-2013-0831.yml
peterhoward45: Fixed.

cves/CVE-2013-0831.yml
peterhoward45: Fixed.

cves/CVE-2013-0831.yml
peterhoward45: Good catch, I added more detail to the file. But basically the fix was changing and an else to an else if.

cves/CVE-2015-6781.yml
peterhoward45: I added more detail to the file. SFNT is a file for font format and SFTNly is library to use it.

cves/CVE-2015-6781.yml
peterhoward45: Good catch! Added more detail about the complexity of font files.

cves/CVE-2015-6781.yml
```diff
@@ -47,26 +49,31 @@ reviews:
...
+    unobtainable.
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
   For the second round of reviewing, you will be giving a certain amount of
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes:
```
michaelboc: +3 Votes

PR #199, opened 2018-10-26
CVE-2010-1505 and CVE-2014-3191

CatherineOsadciw:
Completed analyses of these two CVEs for Round 1.

### DIFF
--- a/cves/CVE-2010-1505.yml
+++ b/cves/CVE-2010-1505.yml
@@ -11,7 +11,7 @@ curated_instructions: |
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
--- a/cves/CVE-2010-1505.yml
+++ b/cves/CVE-2010-1505.yml
@@ -4,7 +4,7 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 'CWE-362: Concurrent Execution using Shared Resource with Improper Synchronization (Race Condition)'
+CWE: 362
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
--- a/cves/CVE-2010-1505.yml
+++ b/cves/CVE-2010-1505.yml
@@ -38,7 +38,10 @@ description: |
   This resulted in the newly-loaded tab sharing the privileges of the new tab page.  
   This vulnerability was the underlying cause of another bug (Chromium bug 12718) 
   which saw form submissions being submitted as a GET instead of a POST since the 
-  page was loaded in the new tab process.
+  page was loaded in the new tab process.  With the privileges of the new tab process,
+  an attacker may have been able to execute a CSRF or similar attack taking advantage of
+  the side effects of these privileges (such as the change of form submissions from POST 
+  to GET).
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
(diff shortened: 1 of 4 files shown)


KevKode:
line# - comment

### CVE-2010-1505
#### Grammar/Spelling/Style:

65 - I'm not sure if the note section of the commit is supposed to
be for the commit message. I think its more for a slight description of 
how that commit contributed the vulnerability or fixed it in your own words.
If not, ignore this.

110 - Is the ie. section an example of how one would implement a unit test?
If so, it is a little unclear to me what steps this test is taking to check
for the vulnerability. I'd reword this to 'An example unit test would:' and 
then list the steps out sentence by sentence.

#### Methodology/Investigation:

114 - What are EXPECT statements, what are they doing? What is 'that variable'? I get that 
a var gets set after NTP is done and checked on pageload, but it took a 
couple read throughs of that sentence to understand.

286 - I really liked how you looped all of the interesting refactoring commits 
into the conclusion. I agree that it seems their shifting was a big contributor to this.

36 - Other than the bug mentioned, is there another way that this vulnerability 
can be exploited that would benefit this description?

137 - What did the existing bug report entail, what was the actual bug,
how did it relate back to this vulnerability?

General - There isn't much on what exploits or bad things that would happen
if this vulnerability was exploited. What would the different permissions give the
attacker access to? How can they use a GET request instead of a POST to attack a site?
Could it be used for CSRF?

General - I'm impressed with the amount of looking around you did to triple check
you had all the right VCCs. Nice!

#### Upvotes: 3

### CVE-2014-3191
#### Grammar/Spelling/Style:

47 - Change the date to YYYY-MM-DD format without the daytime or timezone.

149 - By the answer above it sounds like this vulnerability was discovered 
by an automated program, ClusterFuzz. Might want to swap this to true.

#### Methodology/Investigation:

36 - Is there another way that this vulnerability can be exploited that would 
benefit this description? Honestly there might not be one, to me the description
reads as if updateWidgetPositions is calling unreachable code which would cause a crash, 
use after free bugs, or some other failure.

62 - I'm not sure if the note section of the commit is supposed to
be for the commit message. I think its more for a slight description of 
how that commit contributed the vulnerability or fixed it in your own words.
If not, ignore this.

272 - I like your comments on how the developers decided to fix a little and then fix
everything a year later. Gives an interesting window into the habits of developers.

#### Upvotes: 2

CatherineOsadciw:
Thanks for the feedback!  I fixed all of the commit notes to be small summaries instead of the actual commit messages, as well as added/clarified detail in the test section of 2010-1505 and the description of 2014-3191.

cves/CVE-2014-3191.yml
```diff
@@ -154,38 +192,46 @@ lessons:
...
   complex_inputs:
-    applies: 
+    applies: false
     note: 
+  use_after_free:
+    applies: true
+    note: |
+      This vulnerability was similar to the double-free weakness discussed in class.  Instead of trying to free an already-freed
+      memory location, the code tried to make additional method calls with an object that had already been freed.  Like double-free,
+      the mitigation involves careful memory management and the weakness can easily be found by a fuzzer, as it was here.
```
mesternefeld: Love this description, but take out the class reference. This is going to researches who may or may not have taken Meneely's class.
```suggestion
      This vulnerability is similar to a double-free weakness.  Instead of trying to free an already-freed
```

cves/CVE-2010-1505.yml
```diff
@@ -129,12 +161,38 @@ interesting_commits:
...
+      This was a fairly large refactoring effort for the RenderViewHost, which was used to
+      render the new tabs in this vulnerability.  While this refactoring may not have
+      contributed directly to the VCC, it is very large (1645 insertions and 1226 deletions),
+      which may be a red flag about the size of the commits developers were making in 2008.  
+      If commits of this size were commonplace, it would be much more difficult to locate the root
+      of a vulnerability.
+      
+      This commit was actually the VCC, but the commit message was misleading because it indicated 
+      that it was only a refactoring.  Most of it was indeed refactoring, but buried within all of 
+      those changes was the addition of logic to check if a destination URL already had a SiteInstance
```
BrodrickS: This is very cool observation.
+ 4 upvotes

andymeneely:
Excellent work!

PR #205, opened 2018-10-26
CVE-2010-2108 and CVE-2013-0841

MattThyng:
Filled out files regarding CVE-2010-2108 and CVE-2013-0841

### DIFF
--- a/cves/CVE-2010-2108.yml
+++ b/cves/CVE-2010-2108.yml
@@ -125,8 +125,8 @@ subsystem:
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
   answer: |
-    This was in the chrome subsystem. This is because it is within the chrome directory.
-  name: chrome
+    This was in the browser subsystem. This is because it is within the chrome directory.
+  name: browser
 interesting_commits:
   question: |
     Are there any interesting commits between your VCC(s) and fix(es)?
@@ -155,7 +155,9 @@ major_events:
     we want to capture what the development team was dealing with at the time.
   answer: |
     It appears that at the time the vulnerability was created, the team was focused
-    on updating chrome for the new OS X operating system for mac computers.
+    on updating chrome for the new OS X operating system for mac computers. Even
+    though the vulnerability was introduced while updating for OS X, the
+    vulnerability is for all operating systems.
   events:
   - name: Allowing GPU processing on OS X Mac computers.
     date: 2010-2-3
@@ -225,7 +227,7 @@ mistakes:
     engineering industry would find interesting.
   answer: |
     I think a few mistakes went into the creation of the vulnerability. First
-    of all, I believe their may have been miscommunication within the team. 
+    of all, I believe there may have been miscommunication within the team. 
     When reading the thread detailing the vulnerability, there seemed to be 
     disagreement on whether this was a UI decision. It's possible that when
     the vulnerability was first introduced, the programmer considered it a 
--- a/cves/CVE-2010-2108.yml
+++ b/cves/CVE-2010-2108.yml
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: 284
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: False
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
--- a/cves/CVE-2010-2108.yml
+++ b/cves/CVE-2010-2108.yml
@@ -11,7 +11,7 @@ curated_instructions: |
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: False
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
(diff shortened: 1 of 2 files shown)


cves/CVE-2010-2108.yml
```diff
@@ -11,7 +11,7 @@ curated_instructions: |
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: False
```
andymeneely: MUST be true for this to be graded properly.

MattThyng: Oops, I have fixed it. Curated is not true

cves/CVE-2010-2108.yml
```diff
@@ -195,6 +221,22 @@ mistakes:
     they have written there. Are they doing those? Does the fix look proper?
 
     Use those questions to inspire your answer. Don't feel obligated to answer
-    every one. Write a thoughtful entry here that those ing the software
+    every one. Write a thoughtful entry here that those in the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    I think a few mistakes went into the creation of the vulnerability. First
+    of all, I believe their may have been miscommunication within the team.
```
mesternefeld:
```suggestion
    of all, I believe there may have been miscommunication within the team. 
```

cves/CVE-2013-0841.yml
```diff
@@ -95,10 +111,13 @@ discovered:
...
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
-  date: 
-  automated: 
-  google: 
+  answer: |
+    This vulnerability was discovered by a user and reported to the development team.
+    The vulnerability could be shown when a patch put a multiplier on the index. Then
+    if the user navigated to www.pandora.com, the browser crashed.
```
mesternefeld: Is this specific to www.pandora.com or could we generalize this more? If it is for pandora, that is very interesting.

cves/CVE-2010-2108.yml
```diff
@@ -106,8 +124,9 @@ subsystem:
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
-  answer: 
-  name: 
+  answer: |
+    This was in the chrome subsystem. This is because it is within the chrome directory.
+  name: chrome
```
johnbyrneRIT: Chrome seems a little broad? I could be totally off here, but is there not something like a tabs subsystem?

cves/CVE-2010-2108.yml
```diff
@@ -131,12 +153,14 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer: |
+    It appears that at the time the vulnerability was created, the team was focused
+    on updating chrome for the new OS X operating system for mac computers.
```
johnbyrneRIT: So is the vulnerability Mac specific, or did the changes made for Mac affect other operating systems too?

cves/CVE-2013-0841.yml
```diff
@@ -198,4 +229,12 @@ mistakes:
...
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    This vulnerability was introduced during the implementation phase of development.
+    I believe this vulnerability was the result of a coding mistake, but it could be
+    the result of a design mistake as well. Due to the relatively simple nature of
+    this vulnerability I believe that it could have simply been a mistake done by the
+    programmer. Not properly checking array indexing is something can be easily
+    forgotten. If input validation was not part of the design, then this is a design
+    mistake as well. In order to prevent this vulnerability going forward, it must be
```
johnbyrneRIT: I think the fact that this was reported by a user is also notable. Array index checking is something that unit tests can usually catch, so it makes me wonder if the testing part of the development process is another area for improvement.

cves/CVE-2013-0841.yml
```diff
@@ -32,7 +32,15 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  In the content-blocking funtionality of Google Chrome, there is a
```
johnbyrneRIT: I'm not totally sure what the chrome specific content-blocking functionality is. Is that pop-up blocking? Certificate verification?

cves/CVE-2013-0841.yml
```diff
@@ -77,9 +92,10 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
+    There is no evidence of automated unit tests for this vulnerability.
```
johnbyrneRIT: I would like to give 3 upvotes to this CVE. It's kind of surprising that something like index checking made it all the way to the product.

cves/CVE-2013-0841.yml
MattThyng: I do not believe it is specific for www.pandora.com, but www.pandora.com was able to create the specific environment where this bug happens. I'll not that.

cves/CVE-2010-2108.yml
MattThyng: The change was not specific to Mac, the VCC was only created when they were updating for OS X

cves/CVE-2013-0841.yml
MattThyng: I'll note that

cves/CVE-2013-0841.yml
MattThyng: Well, it is the content-blocking of plugins, pop-ups and more

MattThyng: I'll be more specific

cves/CVE-2013-0841.yml
MattThyng: I'll add that

cves/CVE-2010-2108.yml
MattThyng: I put chrome as the subsystem because of the file structure. I think "browser" may be a better fit

PR #218, opened 2018-10-26
Completed 2011-1813 and 2016-1633

Dormio:
Completed CVE-2011-1813 and CVE-2016-1633

### DIFF
--- a/cves/CVE-2011-1813.yml
+++ b/cves/CVE-2011-1813.yml
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: CWE-825 
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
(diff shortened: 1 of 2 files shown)


cves/CVE-2011-1813.yml
```diff
@@ -94,9 +88,27 @@ discovered:
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
   answer: 
-  date: 
-  automated: 
-  google: 
+    The bug was found by a google employee when examining a testcase.
+    Testcase::
```
andymeneely: Looks like this part broke the yaml. 

And it would fail a future test anyway because `google` is supposed to be just true or false. The longform answer would go in "answer"

While a code snippet is fine, I would suggest summarizing the code snippet as best you can. Your audience will be familiar with programming, but not necessarily with these kinds of subtleties. If you want, you can embed a link to the original snippet too.

cves/CVE-2016-1633.yml
```diff
@@ -76,8 +84,35 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
+  answer: Yes
   code: 
+        <svg width="100%" height="100%" viewBox="0 0 100 100"
```
andymeneely: Same here - this should be yes/no, then the long stuff goes into the "answer". But, it's also a writing assignment, so I would do less copy-paste and more writing to distill down what happened.

cves/CVE-2016-1633.yml
```diff
@@ -198,3 +241,9 @@ mistakes:
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
   answer: 
+    When looking at what was changed in order to fix the CVE, there is just two lines that are deleted
+    and then checks performed to do the same actions that were performed in the line before.
```
andymeneely: Don't just think of this as a coding mistake, think of it as an engineering mistake. What should have happened to prevent this beyond a line or two being wrong?

PR #224, opened 2018-10-28
Kimberly Sookoo -- CVE-2011-3094 and CVE-2014-1701

MukeytheMuse:
Listed in the title are the two (2) CVEs I've selected to investigate for this round, Round 1. Though I did run into some hiccups, I was successfully able to fill out all the answers to the following: CVE description, fix and vcc commits, whether the CVEs were unit tested or not, how the CVEs were initially discovered and when, what subsystems were the mistakes found in, if there were any interesting commits or major events occurring, if these vulnerabilities shared any common lessons from what we've learned so far in class, and what mistakes led to the vulnerabilities being introduced. The only thing left unanswered is the upvotes number, which is not required for Round 1.

### DIFF
--- a/cves/CVE-2011-3094.yml
+++ b/cves/CVE-2011-3094.yml
@@ -119,6 +119,9 @@ discovered:
   answer: |
     According to the information provided at https://www.securitytracker.com/id?1027067,
     a user by the name of "miaubiz" (also known as Brett Wilson) first reported this vulnerability.
+    They explained that they made the initial discovery of this vulnerability when they noticed
+    that a Tibetan letter was followed by a Tibetan cantillation sign, a global buffer overflow
+    occurred.
   date: 2012-04-08
   automated: false
   google: false
--- a/cves/CVE-2011-3094.yml
+++ b/cves/CVE-2011-3094.yml
@@ -162,16 +162,11 @@ major_events:
     we want to capture what the development team was dealing with at the time.
   answer: I saw one major event during this time.
   events:
-  - name: |
-    The event entailed fixing the QtWebKit build with OpenGLES after GC3D/E3D had been
-    refactored. This included the removal of E3DQt and it being replaced with 
-    E3DOpenGLES; enabling defines fot Qt; the addition of pure virtual stubs; the
-    removal of the PROC suffix; the addition of a previously removed function; use of 
-    a port-specific PLATFORM(BLACKBERRY) guard; and the use of USE(OPENGL_ES_2) guard
-    instead of having to always load OpenGL extensions.
+  - name:
+    Fixing the QtWebKit build with OpenGLES after GC3D/E3D had been refactored.
     date: 2012-07-13
   - name: 
-    date: 
+    date:
 lessons:
   question: |
     Are there any common lessons we have learned from class that apply to this
--- a/cves/CVE-2011-3094.yml
+++ b/cves/CVE-2011-3094.yml
@@ -4,7 +4,7 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: 20
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
--- a/cves/CVE-2011-3094.yml
+++ b/cves/CVE-2011-3094.yml
@@ -42,8 +42,9 @@ description: |
   This vulnerability could be exploited by remote attackers, who could execute arbitrary
   code on the system and cause a denial of service via the out-of-bounds read through 
   the use of unspecified vectors. They can accomplish this by persuading a victim to
-  visit a specially-crafted site where they could exploit the vulnerability to trigger the
-  out-of-bounds read in order to execute the arbitrary code on the system.
+  visit a specially-crafted web site or JavaScript program using Chromium or V8, where 
+  they could exploit the vulnerability to trigger the out-of-bounds read in order to 
+  execute the arbitrary code on the system.
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
@@ -52,7 +53,7 @@ bounty:
   date: 
   amount: 0.0
   references:
-    - https://bugs.chromium.org/p/chromium/issues/detail?id=342618
+    - https://bugs.chromium.org/p/chromium/issues/detail?id=122586
 reviews:
 - 10024052
 - 10267007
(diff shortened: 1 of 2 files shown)


cves/CVE-2011-3094.yml
```diff
@@ -32,15 +32,27 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description:
+  The input validation vulnerability did not properly handle Tibetan text, and thus an
```
andymeneely: The reason the CI is failing is from here - doesn't follow the YML syntax for multiline strings. You'll need a pipe character. 

Don't forget that getting the CI to build for Part 1 is a big part of your grade.

cves/CVE-2011-3094.yml
```diff
@@ -32,15 +32,27 @@ description_instructions: |
...
+  out-of-bounds read could occur during Tibetan handling. If a Tibetan letter was followed
+  by a Tibetan cantillation sign, a global buffer overflow would be reported.
+
+  The vulnerability was located in the method called tibetan_form().
+
+  This vulnerability could be exploited by remote attackers, who could execute arbitrary
```
andymeneely: Excellent description, by the way.

cves/CVE-2011-3094.yml
MukeytheMuse: I hadn't realized I missed so many of them. Thank you for bringing it to my attention, Professor. I've fixed them now.

cves/CVE-2011-3094.yml
MukeytheMuse: Thank you :)

cves/CVE-2011-3094.yml
```diff
@@ -32,15 +32,27 @@ description_instructions: |
...
+  code on the system and cause a denial of service via the out-of-bounds read through 
+  the use of unspecified vectors. They can accomplish this by persuading a victim to
+  visit a specially-crafted site where they could exploit the vulnerability to trigger the
+  out-of-bounds read in order to execute the arbitrary code on the system.
```
jmg9728: Would the out-of-bounds read be at risk of occurring if an arbitrary code execution vulnerability existed and was actively leveraging this vulnerability or did the outcome of exploiting this allow for arbitrary code execution? Which came first in this case? I interpreted this as exploiting this out of bounds read vulnerability allows for an arbitrary code execution vulnerability to exist. If that is true then maybe more information into how that is possible would be interesting to read about.

cves/CVE-2011-3094.yml
```diff
@@ -197,4 +231,9 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    A mistake that allowed for that out-of-bounds error was that some variable "c", which
+    most likely represents the index for the buffer, was allowed to be either less than or
+    equal to 0x0fc0, which in turn allowed that out-of-bounds error to occur. It appeared
+    to have been an oversight in how the boundary cases were addressed and thus limited
+    incorrectly.
```
jmg9728: "c" in this case represents the Tibetan unicode char hex value and not an index location, this is however used with the math that gives you an index location. In this case tibetanForm[] allocates 128 indexes and the range of accepted unicode chars was 129 [0x0f40,0x0fc0] thus allowing an out of bounds read up to a single index past what was allocated. To get the desired offset in the array, "c" was subtracted from 0x0f40 to give offsets of 0-127. Once the gating on allowed boundaries was updated in the fix commit this ensured only index values 0-127 and not 0-128.

cves/CVE-2011-3094.yml
```diff
@@ -32,15 +32,27 @@ description_instructions: |
...
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
   was wrong. Otherwise, leave it blank.
 bounty:
   date: 
-  amount: 
-  references: []
+  amount: 0.0
+  references:
+    - https://bugs.chromium.org/p/chromium/issues/detail?id=342618
```
NicholasFeldman: This reference doesn't appear to be right, it look like [this](https://bugs.chromium.org/p/chromium/issues/detail?id=122586) is the bug report and [these](https://chromereleases.googleblog.com/2012/05/stable-channel-update.html) are the release notes

cves/CVE-2011-3094.yml
```diff
@@ -156,17 +185,22 @@ lessons:
     applies: 
     note: 
   least_privilege:
-    applies: 
-    note: 
+    applies: true
+    note: |
+      The attacker would only need the same privilege as an average user.
```
NicholasFeldman: Can you elaborate on this a bit more? How was the attacker given more privilege than an average user?

cves/CVE-2014-1701.yml
```diff
@@ -95,10 +129,13 @@ discovered:
...
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
-  date: 
-  automated: 
-  google: 
+  answer: |
+    According to the information provided at https://www.debian.org/security/2014/dsa-2883,
+    a user by the name "aidanhs" had been the first one to discover a cross-site scripting
+    issue in event handling.
```
NicholasFeldman: Is there any insight as to how this vulnerability was discovered? More information might be available from the [bug report](https://bugs.chromium.org/p/chromium/issues/detail?id=342618)

cves/CVE-2011-3094.yml
```diff
@@ -119,8 +145,11 @@ interesting_commits:
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
   answer:
   commits:
-  - commit: 
-    note: 
+  - commit: e161ba15b5be329b35351b6242d61852bfd2b4b1
+    note: |
+      A fix was implemented to prevent app banners when a matching hosted/bookmark app
```
bencrossgrove: Can you explain what app banners are?

cves/CVE-2011-3094.yml
MukeytheMuse: You're absolutely correct. I'd added a bunch of references before as I misinterpreted what I was citing them for, and it looks like I kept the wrong reference by mistake. Thank you for pointing that out.

cves/CVE-2011-3094.yml
MukeytheMuse: Thank you for explaining this to me! I severely misunderstood the explanation for what "c" actually represented!

cves/CVE-2011-3094.yml
MukeytheMuse: My apologies, I had some trouble logging these commits the first time around, and it turns out, this commit is unrelated entirely and not part of the log between the fix and the corrected VCC.

MukeytheMuse: Also an app banner is basically that box that appears on the top of your screen on mobile that tells you that you should use the native app if it's available rather than access the content through the web.

cves/CVE-2011-3094.yml
MukeytheMuse: I actually didn't consider that they'd need to be given more privilege than an average user at first. What I understood from this vulnerability was that the attacker only needed to deceive the victim and execute some arbitrary code in order to trigger the out-of-bounds read. They're basically able to cause a DoS attack, so I believe that would count to some extent as also having greater privilege.

cves/CVE-2014-1701.yml
MukeytheMuse: Thank you for pointing this out to me! I'll make sure to add more details about its discovery.

PR #233, opened 2018-10-28
CVE-2010-0644 and CVE-2014-9689

bjpicciano:
Round 1 Submission

### DIFF
--- a/cves/CVE-2010-0644.yml
+++ b/cves/CVE-2010-0644.yml
@@ -72,7 +72,7 @@ upvotes_instructions: |
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes: 4
 unit_tested:
   question: |
     Were automated unit tests involved in this vulnerability?
--- a/cves/CVE-2010-0644.yml
+++ b/cves/CVE-2010-0644.yml
@@ -4,21 +4,21 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: 200
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
   source for this is Chrome's Stable Release Channel
   (https://chromereleases.googleblog.com/).
   Please enter your date in YYYY-MM-DD format.
-announced: 2010-02-18 13:00:00.410000000 -05:00
+announced: 2009-12-09
 description_instructions: |
   You can get an initial description from the CVE entry on cve.mitre.org. These
   descriptions are a fine start, but they can be kind of jargony.
@@ -32,7 +32,12 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  In Google Chrome when a SOCKS 5 proxy server is configured, DNS queries should
+  be forwarded through a SOCKS 5 proxy. In this specific version, DNS queries
+  were performed from the local network instead. This could allow for remote DNS
+  servers to obtain sensitive information about the identity of the client user
+  via requets logging.
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
@@ -54,12 +59,12 @@ fixes_vcc_instructions: |
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: f209dba88c91b0461dc5acd02af0b3685213938c
-  :note: ''
+  :note: The fix made it so the client-side DNS resolution defaulted to proxy-side (instead of local).
 - :commit: 20cbe23d256ad1522424a030fd109b82c7d08938
-  :note: ''
+  :note: Removed the support for IPv4 and IPv6 addressing types in the SOCKS 5 implementation.
 vccs:
-- :commit:
-  :note: 
+- :commit: e0c27beabdf16ead25bfba3b7db2965efef19ee1
+  :note: this was introduced when SOCKS 5 was implementated into Chrome.
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
@@ -80,9 +85,9 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: A new test was created
+  code: true
+  fix: true
 discovered:
   question: |
     How was this vulnerability discovered?
@@ -98,11 +103,12 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: | 
+    It was discovered by a user. They noticied through wireshark they were sending their ip address instead of going through the proxy.
+  date: 2009-12-09
+  automated: false
+  google: false
+  contest: false
 subsystem:
   question: |
     What subsystems was the mistake in?
@@ -110,8 +116,8 @@ subsystem:
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
-  answer: 
-  name: 
+  answer: Based on files altered to fix.
+  name: net
 interesting_commits:
   question: |
     Are there any interesting commits between your VCC(s) and fix(es)?
@@ -121,7 +127,8 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer:
+  answer: | 
+    I didn't find any interesting commits. Many dealt with mac support and minor bug fixes. They also implemented visulizations with histograms.
   commits:
   - commit: 
     note: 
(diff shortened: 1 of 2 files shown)


cves/CVE-2010-0644.yml
```diff
@@ -80,9 +85,9 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: A new test was created
```
matthewD-AVI: Perhaps a brief explanation of the test

cves/CVE-2010-0644.yml
```diff
@@ -32,7 +32,12 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  In Google Chrome when a SOCKS 5 proxy server is configured, DNS queries should
+  be forwarded through a SOCKS 5 proxy. In this specific version, DNS queries
+  were performed from the local network instead. This could allow for remote DNS
+  servers to obtain sensitive information about the identity of the client user
+  via requets logging.
```
ZacharyMoran: This is extremely close to the actual description on the NVD site, not that this is a bad thing, but it is still missing the two things that I am left questioning with both your description and the NVD one:

1. How does the remote DNS server obtain the info.
2. Any minor description of a SOCKS 5 proxy server. Maybe I'm just not informed well but I have no idea what this is nor do I expect "anyone with some programming experience" to necessarily know what this is.

cves/CVE-2010-0644.yml
ZacharyMoran: Agree with the above, some information regarding what the test was looking for, how it was implemented, or what the approach of the test would point out if it were to fail.

cves/CVE-2010-0644.yml
```diff
@@ -4,21 +4,21 @@ CWE_instructions: |
...
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
   source for this is Chrome's Stable Release Channel
   (https://chromereleases.googleblog.com/).
   Please enter your date in YYYY-MM-DD format.
-announced: 2010-02-18 13:00:00.410000000 -05:00
+announced: 2009-12-09
```
ZacharyMoran: This is the date that the bug was reported / found by the user. The date intended to go here is the original one (or close to it) as this is asking for when the bug was released to the world. The Chrome Releases blogpost with a publicized version of this bug was on Wednesday, February 10, 2010 (https://chromereleases.googleblog.com/2010/02/stable-channel-update.html)

cves/CVE-2010-0644.yml
```diff
@@ -4,21 +4,21 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: 200
```
ZacharyMoran: +2 upvotes

cves/CVE-2014-9689.yml
```diff
@@ -4,21 +4,21 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: 264
```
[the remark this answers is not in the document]
ZacharyMoran: I feel like CWE-276 could apply a bit more focused efforts to this CVE. (https://cwe.mitre.org/data/definitions/276.html). This would also give you room to discuss the mitigations in the CWE rather than just being stuck with a CWE Category classification.

cves/CVE-2014-9689.yml
```diff
@@ -4,21 +4,21 @@ CWE_instructions: |
...
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
   source for this is Chrome's Stable Release Channel
   (https://chromereleases.googleblog.com/).
   Please enter your date in YYYY-MM-DD format.
-announced: 2015-03-08 20:59:02.450000000 -04:00
+announced: 2014-10-09
```
ZacharyMoran: Again similar to the previous one:
This is the date that the bug was reported / found by the user. The date intended to go here is the original one (or close to it) as this is asking for when the bug was released to the world. The Chrome Releases blogpost with a publicized version of this bug was on Tuesday, March 3, 2015 (https://chromereleases.googleblog.com/2015/03/stable-channel-update.html)

cves/CVE-2014-9689.yml
```diff
@@ -32,7 +32,11 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+    Chrome was exposing the accelerometer and gyroscope data to websites without
+    asking for permission on iOS and Andriod. This made it easier for attackers monitor keystrokes,
+    track users, and even obtain low frequency speech signals. Current technology in mobile devices
+    do not allow for such precise measurements, but it is technically possible.
```
ZacharyMoran: How did accelerometer and gyroscopic data allow for keylogging? I'm not quite understanding the connection.

cves/CVE-2014-9689.yml
```diff
@@ -51,10 +55,10 @@ fixes_vcc_instructions: |
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: d771a782774564197c9c4d3ddd07605e53397a12
-  :note: ''
+  :note: The fix involved creating a test to make sure the sensors do not go above 60hz.
 vccs:
-- :commit:
-  :note: 
+- :commit:
```
ZacharyMoran: Were you unable to find a commit or was it outside of the scope of the chromium repository? Because regardless of whether or not the sensors are present, implementation to use them has to be in place and that implementation would have been where the VCC came from as it would have been lacking the permission check.

cves/CVE-2014-9689.yml
```diff
@@ -130,7 +139,7 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer: No major events happend at this time.
```
ZacharyMoran: If there is no major events, what went on during the time between the initial implementation of the sensors and the VCC? (May need to actually identify a VCC to address this.)

cves/CVE-2014-9689.yml
```diff
@@ -179,8 +188,8 @@ lessons:
     applies: 
     note:
```
ZacharyMoran: Potentially could apply, I can see how they just did not assume the sensitivity of the sensors would result in a vulnerability like this.

cves/CVE-2014-9689.yml
```diff
@@ -196,4 +205,7 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    The mistake they made was not realising the types of vulnerabilities created by giving anyone access
+    to a user's extremely percise gyroscope and accelerometer. That level of precision is not needed in the
+    vast majority of usecases and if it was, it should be a permission that the user can give.
```
ZacharyMoran: Did they apply any of the mitigation strategies listed on CWE-276 if you decide to change your classification with my recommendation. If you choose not to change it what generally acceptable permissions mitigations did they utilize?

cves/CVE-2014-9689.yml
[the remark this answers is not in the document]
bjpicciano: This is an interesting point, but this CVE isn't entirely related to the permissions error but rather the precision of the instruments leading to unwanted information being discoverable. The permissions side was just a potential fix to this theoretical problem

cves/CVE-2014-9689.yml
bjpicciano: Ill rewrite the description clearer with less emphasis on permissions.

cves/CVE-2014-9689.yml
bjpicciano: Even though this is a theoretical bug, the VCC would be the inception of the accelerator/gyroscope's instruments?

cves/CVE-2014-9689.yml
bjpicciano: They did not apply any permission mitigations because the device's instruments are not that precise yet. Instead they pushed a testcase and limited the readings to 60hz.

cves/CVE-2010-0644.yml
bjpicciano: Wasn't that the date of the fix being implemented? I guess that was when the majority of the public first new of it.

cves/CVE-2010-0644.yml
bjpicciano: Yes, this test was to ensure the devices only sent signals at 60hz. This is not precise enough to obtain these vulnerabilities

bjpicciano: Yes, the unit test was to make sure the default setting would use the proxy

cves/CVE-2014-9689.yml
bjpicciano: I have updated it to show the date of the paper publicized from Stanford. This is what spurred the user to post the issue, and a variety of new outlets also reported about it.

cves/CVE-2014-9689.yml
bjpicciano: Yes, i'd assume if the instruments needed to be more precise then permission would be necessary.

cves/CVE-2010-0644.yml
```diff
@@ -54,12 +59,12 @@ fixes_vcc_instructions: |
...
 - :commit: 20cbe23d256ad1522424a030fd109b82c7d08938
-  :note: ''
+  :note: Removed the support for IPv4 and IPv6 addressing types in the SOCKS 5 implementation.
 vccs:
-- :commit:
-  :note: 
+- :commit: e0c27beabdf16ead25bfba3b7db2965efef19ee1
+  :note: this was introduced when SOCKS 5 was implementated into Chrome.
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
```
sxHelloWorld: +2 upvotes

cves/CVE-2010-0644.yml
bjpicciano: So when the user is trying to obtain the IP address from the Domain name so they can connect to the site, they send a request to the DNS. The DNS sees the IP address from which it comes from and can log it. Some users wish to send their request through the SOCKS 5 proxy so the DNS can not see the users IP address. This also prevents onlookers from these queries as well.

PR #236, opened 2018-10-29
CVE 2011-1434 and CVE 2015-6769

Maxzolythus:
Completed Part 1

### DIFF
--- a/cves/CVE-2011-1434.yml
+++ b/cves/CVE-2011-1434.yml
@@ -92,8 +92,8 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer:  
-  code: It did have a test case, it had a browser test that found several files which triggered this fairly often.
+  answer: It did have a test case, it had a browser test that found several files which triggered this fairly often.
+  code: true
   fix: They did not change the tests
 discovered:
   question: |
--- a/cves/CVE-2011-1434.yml
+++ b/cves/CVE-2011-1434.yml
@@ -79,7 +79,7 @@ upvotes_instructions: |
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes: 3
 unit_tested:
   question: |
     Were automated unit tests involved in this vulnerability?
--- a/cves/CVE-2011-1434.yml
+++ b/cves/CVE-2011-1434.yml
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: 20
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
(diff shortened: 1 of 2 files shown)


cves/CVE-2011-1434.yml
```diff
@@ -153,34 +173,34 @@ lessons:
...
-    applies: 
+    applies: false
     note: 
-  secure_by_default:
-    applies: 
+  secure_by_default: 
+    applies: false
     note: 
-  yagni:
+  yagni: false
```
andymeneely: leave this line empty, and put the false into "applies". That's why this is failing CI. Don't forget that passing the CI is part of your Part 1 grade.

cves/CVE-2015-6769.yml
```diff
@@ -119,8 +147,9 @@ interesting_commits:
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
   answer:
   commits:
-  - commit: 
-    note: 
+  - commit: 63c9596d206c896b35ae8a5bff4db73847f62f33
+    note: |
+      They were fixing other stuff. But the thing they were fixing has large parts of this code rewritten.
```
jak5779: This should be expanded. In particular: what was being fixed, how did that change code nearby the issue, and what was the impact of that change?

cves/CVE-2011-1434.yml
```diff
@@ -77,9 +92,9 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: It did have a test case, it had a browser test that found several files which triggered this fairly often.
```
jak5779: Was that an automated unit test, though? That isn't clear, especially since it seems to have been found by someone running (beta) production code - how did that get released if it was failing the automated tests?

cves/CVE-2015-6769.yml
```diff
@@ -153,37 +182,37 @@ lessons:
...
     free to give it a small name and add one in the same format as these.
   defense_in_depth:
-    applies: 
+    applies: false
     note: 
   least_privilege:
-    applies: 
-    note: 
+    applies: true
+    note: If the user didn't have to abillity to execute actions during the unload step then this vulnerability wouldn't have existed
```
jak5779: That is... not a terribly helpful perspective. Cleanup is a very important task which can't really be ignored like that.

jak5779:
I'd give CVE-2011-1434 three upvotes. While the issue itself wasn't especially unique (multi-threaded code is a constant source of headaches for programmers everywhere...), the process by which it was resolved was somewhat interesting. Especially so was the large amount of data that was produced and analyzed over the bug's lifespan.

I'd give CVE-2015-6769 no upvotes, since it simply involved the shutdown procedure having been ordered slightly incorrectly. An important issue to fix, certainly, but not a very interesting one.

cves/CVE-2011-1434.yml
Maxzolythus: Great question, it was found in beta but also the current dev at the time. From the comments in the bug report it seems like it was automated. It's possible that they were either "lucky" when testing, because only one test seemed to catch it. Or that the files have been changed like one who commented mentioned.

cves/CVE-2015-6769.yml
Maxzolythus: Clean up is important. But that wasn't the main effort of this CVE, at the core this script inadvertently allowed you to elevate your privilege and bypass parts of execution. Which should have not been allowed in the first place.

cves/CVE-2015-6769.yml
Maxzolythus: Thanks for the feedback, this was a place holder that I forgot to remove. Fixing that now.

PR #237, opened 2018-10-29
CVE-2011-1059 and CVE-2011-2845

tjb4578:
Curated CVE-2011-1059 and CVE-2011-2845

### DIFF
--- a/cves/CVE-2011-1059.yml
+++ b/cves/CVE-2011-1059.yml
@@ -33,6 +33,9 @@ description_instructions: |
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
 description: 
+  A vulnerability that allows remote attackers the ability cause a denial of 
+  service by enticing a user to resubmit a form.
+
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
@@ -51,10 +54,10 @@ fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
-- :commit: ea889ae0f3e6fb9902b5be7c5dd9ff8fa0e522e9
-  :note: ''
+- :commit: 29660d185211a9823bd9ebbfe147282f1b23e67b
+  :note: This fix came about 1 month after the vulnerability was introduced.
 vccs:
-- :commit:
+- :commit: 71075a2dc134ffd5d33a8d140fb21121f6832764
   :note: 
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
@@ -77,8 +80,10 @@ unit_tested:
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
   answer: 
-  code: 
-  fix: 
+    Yes, from the code provided in the fix, it was clear that there were tests
+    written both before and after the fix.
+  code: True
+  fix: True
 discovered:
   question: |
     How was this vulnerability discovered?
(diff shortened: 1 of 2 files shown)


cves/CVE-2011-1059.yml
```diff
@@ -197,4 +210,8 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: This was a simple case of trying to access memory that has been freed.  
+          It only took about 12 days to discover the bug, but it took about 10
+          days to reproduce it because it happens on a rare boundary condition.
+          Every developer makes memory management mistakes like this once in a 
+          while.
```
jacobklssn: Adding a tip on how to possibly avoid this problem would sound better than just saying that it was/is inevitable for every developer to cause a vulnerability like this. Even if it's true.

cves/CVE-2011-1059.yml
```diff
@@ -4,7 +4,7 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
```
jacobklssn: upvotes: 3

cves/CVE-2011-2845.yml
```diff
@@ -207,3 +226,6 @@ mistakes:
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
   answer: 
+    The mistake here was that there wasn't proper communication between
+    the NavigationController and the Renderer.  This created a race condition
+    that confused the NavigationController into allowing stale entries.
```
jacobklssn: Describing NavigationController as what it does instead of its name could help get across your point. I can make some guesses, but I don't have any concrete idea of what NavigationController does.

cves/CVE-2011-2845.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
```
jacobklssn: upvotes: 5

cves/CVE-2011-2845.yml
tjb4578: I updated the description, but I'm still not super clear myself on how the NavigationController, HistoryController, and Renderer interact.  All I could gather from the conversations between the devs was that a race condition occurred when the NavigationController and Renderer became out of sync

cves/CVE-2011-1059.yml
tjb4578: Good idea.  Most security scanners will pick up use-after-free vulnerabilities.  I'll add that in

cves/CVE-2011-1059.yml
```diff
@@ -63,7 +67,7 @@ upvotes_instructions: |
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes: 3
```
devinmatte: Upvotes: 3

cves/CVE-2011-2845.yml
```diff
@@ -72,7 +79,7 @@ upvotes_instructions: |
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes: 5
```
devinmatte: Upvotes: 5

cves/CVE-2011-2845.yml
```diff
@@ -103,20 +111,26 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: This vulnerability was discovered by user jconsult...@gmail.com.  At first, his
```
devinmatte: Are you simply unable to find the rest of the email?

cves/CVE-2011-2845.yml
```diff
@@ -207,3 +226,8 @@ mistakes:
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
   answer: 
+    The mistake here was that there wasn't proper communication between
+    the NavigationController and the Renderer.  Since the NavigationController
+    needs to communicate with the HistoryController to figure out where to go
+    when the user presses the back button, this allowed time for a race condition
+    to form that confused the NavigationController into allowing stale entries.
```
devinmatte: You're using a lot of Chromium and file specific language, are you able to more generalize this issue?

cves/CVE-2011-2845.yml
tjb4578: Yeah it didn't display his full email for some reason

PR #239, opened 2018-10-29
CVE-2013-0843 and CVE-2016-1698

gwenmunson:
Creating pull request for History Part 1, first pull request. May have to redo if syntax isn't right.


mesternefeld:
Please revert file for CVE-2013-6630 since it is not touched.

cves/CVE-2016-1698.yml
```diff
@@ -98,20 +121,28 @@ discovered:
...
   question: |
     What subsystems was the mistake in?
 
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
-  answer: 
-  name: 
+  answer: renderer
+  name: extensions/renderer/resources/binding.js
```
mesternefeld: this isn't the subsystem but rather the file. I believe the subsystem was extensions.

gwenmunson: Thanks for the heads up! I've updated the YAML files and pushed them, so everything should be good.

cves/CVE-2013-0843.yml
```diff
@@ -4,21 +4,21 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: CWE-787
```
Maxzolythus: This is a little nit picky. But from looking through the bug report and code snippets it seems less like they directly write to somewhere out of bounds. However they do seem to not take care in making sure something doesn't write to out of bounds (CWE-119).

cves/CVE-2013-0843.yml
```diff
@@ -32,7 +32,12 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  For Mac OS, the buffer size is too small when running the 96 kHz sampling
```
Maxzolythus: Great description!

cves/CVE-2013-0843.yml
gwenmunson: Thanks for the feedback!

Looking over my work and the CVE, I do agree that it's not necessarily writing out of bounds; other systems have tests for the 96kHz rate, showing that it's not a value that's out of bounds. Even with that said, it doesn't seem like they were aware of where their boundary of values are, considering it was presumed in the code that the MacOSX would not be using a 96 kHz rate (whether this was intentional or an accidental omission isn't 100% clear, although evidence with the other systems point more towards the former). I'll look at editing my wording to not make it seem like 96kHz was out of bounds and talk more about understanding the range of values that a code snippet will be run against.

cves/CVE-2013-0843.yml
```diff
@@ -154,37 +194,47 @@ lessons:
...
     note: 
   secure_by_default:
-    applies: 
-    note: 
+    applies: true
+    note: |
+      It was assumed that the commit that caused the issue would no longer arise in
+      any problems and that this data could not be messed with. It was not assumed
+      that there may be any other issues belonging to MacOS only, and, if this
+      were thought about more in depth, the issue may have been prevented.
```
jak5779: This needs some clarification and rewording. Some questions it raises:
* When was the assumption made? Before the VCC? After the VCC but before the fix? Etc.
* What does it mean for "the commit ... [to] arise in any problems"?
* By "It was not assumed that there may be any other issues belonging to MacOS only", do you mean that the MacOS code was not investigated for vulnerabilities closely enough?
* Who was assumed to have been prevented from having "messed with [this data]"?
* How does this note connect to the concept of ensuring default settings are safe?

cves/CVE-2013-0843.yml
```diff
@@ -198,4 +248,11 @@ mistakes:
...
     engineering industry would find interesting.
-  answer: 
+  answer: |
+   I believe that, since this vulnerability was due to not considering
+   the use of 96 kHz sampling rates for MacOSX, including additional 
+   cases of increased rates (96 kHz, 192 kHz) should be included in the unit
+   testing scripts for all platforms to ensure that this type of vulnerability
+   is not repeated. It's important to have good boundary testing in order to
+   determine how your code works at, in, and outside of your expected value of
+   ranges as it can cause other issue that the system may not be prepared for.
```
jak5779: Great answer! Only a few small grammar oversights to pick out:

>including additional cases of increased rates (96 kHz, 192 kHz) should be included in the unit testing scripts
The first "including" should be removed as redundant.

>for all platforms to ensure that this type of vulnerability
Run-on sentence: put a sentence break after "platforms".

>determine how your code works at, in, and outside of your expected value of ranges
This should be something more like "determine how your code works when values are in, around, and outside of their expected ranges".

>ranges as it can cause other issue that the system may not be prepared for.
There should be a comma after "ranges", "it" should be "that", and "other issue" should just be "issues".

cves/CVE-2016-1698.yml
```diff
@@ -98,20 +121,28 @@ discovered:
...
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: |
+    This vulnerability was discovered by rob@robwu.nl when working on unit tests.
+    Most interestingly enough, Rob had come up with a large proof of concept that
```
jak5779: >Most interestingly enough

Seems two different sentence openers ran into each other here

jak5779:
I'd give CVE-2016-1698 five upvotes due to being such a glaring example of how long problems can stick around, unsuspected, if they are not tested for. This is especially the case here when there wasn't any validation occurring at all for the module type here. Additionally, it's interesting to note that the fix involved hard-coding a whitelist in the code itself. Normally that wouldn't be a great design element, but its use here does raise the point that sometimes it's not worth expending additional effort to strictly follow best-practices.

I'd give CVE-2013-0843 two upvotes, since it is a fairly typical example of unexpected values causing unexpected behavior, but winds up being complicated somewhat by platform differences. This emphasizes the role unit tests play in ensuring platform independence.

cves/CVE-2013-0843.yml
scioffi: The CWE value should just be the CWE ID number.

```suggestion
CWE: 787
```

cves/CVE-2016-1698.yml
```diff
@@ -4,21 +4,21 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: CWE-20
```
scioffi: The CWE value should just be the CWE ID number.
```suggestion
CWE: 20
```

cves/CVE-2013-0843.yml
gwenmunson: Thanks for the feedback!

I do disagree with "for all platforms to ensure that this type of vulnerability" being a run-on sentence; "to ensure that this type of vulnerability" and what follows is still an incomplete sentence clause that lacks a subject. Other than that, I've implemented the changes here and will commit them shortly.

cves/CVE-2016-1698.yml
```diff
@@ -98,20 +121,28 @@ discovered:
...
-  contest: 
+  answer: |
+    This vulnerability was discovered by rob@robwu.nl when working on unit tests.
+    Most interestingly enough, Rob had come up with a large proof of concept that
+    demonstrated exactly how detrimental this vulrenability could be. During testing,
+    he noticed that the schema objects persisted across page loads, allowing it to 
+    be modified when the page is reloaded, thus accessing arbitrary bindings and
+    disclosing information. The bugs page contains a lot of in-depth information
+    about this issue and is quite impressive as it has a high level of polish
+    and testing done, especially for a non-Chrome employee.
```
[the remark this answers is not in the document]
gwenmunson: Thanks for the feedback!

For the last comment, I changed it to "The bug pages" as that made more sense within context. I've implemented these changes and will commit these soon.

cves/CVE-2013-0843.yml
gwenmunson: Thanks for the feedback!

The assumption was made after the VCC but before the fix, and it was believed that the code for MacOSX would not run into any problems as the range of rates it would handle were all tested and covered. By "It was not assumed that there may be any other issues belonging to MacOS only"  I did indeed mean that the MacOSX code was not investigated for vulnerabilities closely enough as, if it were looked at more, this issue could have been prevented since 96 kHz was one of the rates that the buffer couldn't handle at the time. The person who was prevented from messing with the data was any user - malicious or not - that would cause the buffer to overflow, causing memory corruption, crashing the application, and potentially leaking sensitive information to a site that provides WebRTC audio; a malicious attacker could exploit this to cause a denial of service attack. Looking over the scenario a bit more, it doesn't fit as much under secure by default as I thought it did, but my initial reasoning was that the initial settings for MacOSX were assumed to be safe. I'll see about rewording this and maybe moving this to another application in class as it doesn't fit secure by default as much as I thought it did.

PR #240, opened 2018-10-29
CVE-2010-3117 and CVE-2011-0776

jak5779:
For SWEN-331 class.

### DIFF
--- a/cves/CVE-2010-3117.yml
+++ b/cves/CVE-2010-3117.yml
@@ -61,7 +61,7 @@ fixes:
   :note: Shutdown the notification UI manager before the notifications service, as the former's shutdown logic depends on the latter.
 vccs:
 - :commit: e13ad79b6359b68fbefcd589806cf59106cc2ca7
-  :note: Commit message `Continual header pruning`. This is the commit where the notification UI manager had its lifespan moved entirely into browser_process_impl.cc
+  :note: Commit message Continual header pruning. This is the commit where the notification UI manager had its lifespan moved entirely into browser_process_impl.cc
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
--- a/cves/CVE-2010-3117.yml
+++ b/cves/CVE-2010-3117.yml
@@ -4,7 +4,7 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: Other - no entry seems to exist for an incorrectly *ordered* cleanup - the closest only applies to cases where an exception has been raised. Note that the nvd.nist.gov page agrees with this assessment.
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
--- a/cves/CVE-2010-3117.yml
+++ b/cves/CVE-2010-3117.yml
@@ -59,12 +59,10 @@ fixes:
 - :commit: 97452c9788b026b810bd4fbd55cfd40711548e3d
   :note: Reset the notification ui manager even earlier, before the IO thread goes away.
 - :commit: 9c1fe143f8e11092424875fbcdaccec16f141981
-  :note: Shutdown the notification UI manager before the notifications service, as the former's
-            shutdown logic depends on the latter.
+  :note: Shutdown the notification UI manager before the notifications service, as the former's shutdown logic depends on the latter.
 vccs:
 - :commit: e13ad79b6359b68fbefcd589806cf59106cc2ca7
-  :note: Commit message: "Continual header pruning". This is the commit where the notification UI
-            manager had its lifespan moved entirely into browser_process_impl.cc
+  :note: Commit message: "Continual header pruning". This is the commit where the notification UI manager had its lifespan moved entirely into browser_process_impl.cc
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
@@ -126,11 +124,7 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer: Only two commits occurred between the VCC and the first of the fixes. One was a minor
-            change responding to the movement of a header file into a subdirectory, while the other
-            appears to involve a small refactor which changed which implementation of a function was
-            to be used. In other words, neither commit could be considered "interesting" from the
-            perspective of this issue.
+  answer: Only two commits occurred between the VCC and the first of the fixes. One was a minor change responding to the movement of a header file into a subdirectory, while the other appears to involve a small refactor which changed which implementation of a function was to be used. In other words, neither commit could be considered "interesting" from the perspective of this issue.
   commits:
 major_events:
   question: |
@@ -140,8 +134,7 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: The closest thing to a major event was the bug report initially being attached to several
-    unrelated crashes.
+  answer: The closest thing to a major event was the bug report initially being attached to several unrelated crashes.
   events:
   - name: Identified attached crashes as unrelated.
     date: 2010-08-03
@@ -161,37 +154,37 @@ lessons:
     If you think of another lesson we covered in class that applies here, feel
     free to give it a small name and add one in the same format as these.
   defense_in_depth:
-    applies: 
-    note: 
+    applies: true
+    note: A crash due to some other component having been unloaded should not result in a sandbox escape. Instead, it should be caught and handled.
   least_privilege:
-    applies: 
+    applies: false
     note: 
   frameworks_are_optional:
-    applies: 
+    applies: false
     note: 
   native_wrappers:
-    applies: 
+    applies: false
     note: 
   distrust_input:
-    applies: 
+    applies: false
     note: 
   security_by_obscurity:
-    applies: 
+    applies: false
     note: 
   serial_killer:
-    applies: 
+    applies: false 
     note: 
   environment_variables:
-    applies: 
+    applies: false
     note: 
   secure_by_default:
-    applies: 
+    applies: false
     note: 
   yagni:
-    applies: 
+    applies: false
     note: 
   complex_inputs:
-    applies: 
+    applies: false
     note: 
 mistakes:
   question: |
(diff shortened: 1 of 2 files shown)


cves/CVE-2010-3117.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: Other - no entry seems to exist for an incorrectly *ordered* cleanup - the closest only applies to cases where an exception has been raised. Note that the nvd.nist.gov page agrees with this assessment.
```
siberiax: This value should just be an integer. If there was no CWE that seemed appropriate it should be left blank

cves/CVE-2010-3117.yml
```diff
@@ -32,15 +32,15 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: Chrome crashes on shutdown due to the notification UI manager not being closed early enough.
```
siberiax: This description needs to be expanded. Explanation of what specifically is wrong with regards to the code would be useful

cves/CVE-2010-3117.yml
```diff
@@ -123,12 +125,10 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer:
+  answer: Only two commits occurred between the VCC and the first of the fixes. One was a minor change responding to the movement of a header file into a subdirectory, while the other appears to involve a small refactor which changed which implementation of a function was to be used. In other words, neither commit could be considered interesting from the perspective of this issue.
```
siberiax: Good explanation here!

cves/CVE-2011-0776.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: CWE-200
```
siberiax: No need to write "CWE-200". Just an integer is required here. Should be just 200

cves/CVE-2011-0776.yml
```diff
@@ -32,15 +32,15 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: On Mac, Chrome's sandbox can be escaped to obtain information about local files.
```
siberiax: Again, expand more on the description of how this happens

cves/CVE-2011-0776.yml
```diff
@@ -197,4 +200,4 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: The issue ultimately derived from Mac allowing calling stat() on arbitrary paths when that isn't expected. This required Chromium to adjust its access to that call. The issue also could have been avoided if Mac's interface to the stat() system call was in line with user expectations.
```
siberiax: Good answer here!

cves/CVE-2010-3117.yml
```diff
@@ -126,11 +124,7 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer: Only two commits occurred between the VCC and the first of the fixes. One was a minor
-            change responding to the movement of a header file into a subdirectory, while the other
-            appears to involve a small refactor which changed which implementation of a function was
-            to be used. In other words, neither commit could be considered "interesting" from the
-            perspective of this issue.
+  answer: Only two commits occurred between the VCC and the first of the fixes. One was a minor change responding to the movement of a header file into a subdirectory, while the other appears to involve a small refactor which changed which implementation of a function was to be used. In other words, neither commit could be considered "interesting" from the perspective of this issue.
```
Maxzolythus: You can keep this old formatting by sectioning off your text by using the pipe character

cves/CVE-2010-3117.yml
```diff
@@ -100,20 +101,21 @@ discovered:
...
 subsystem:
   question: |
     What subsystems was the mistake in?
 
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
-  answer: 
-  name: 
+  answer: Based on the filename and bug report.
```
gwenmunson: It'd be nice to have the exact data here rather than just pointing to the filename and bug report for a clear, concise answer that anyone can reference when looking at this CVE in the future.

cves/CVE-2010-3117.yml
```diff
@@ -55,13 +55,13 @@ fixes_vcc_instructions: |
...
+- commit: 97452c9788b026b810bd4fbd55cfd40711548e3d
+  note: Reset the notification ui manager even earlier, before the IO thread goes away.
+- commit: 9c1fe143f8e11092424875fbcdaccec16f141981
+  note: Shutdown the notification UI manager before the notifications service, as the former's shutdown logic depends on the latter.
 vccs:
-- :commit:
-  :note: 
+- commit: e13ad79b6359b68fbefcd589806cf59106cc2ca7
+  note: Commit message Continual header pruning. This is the commit where the notification UI manager had its lifespan moved entirely into browser_process_impl.cc
 upvotes_instructions: |
```
gwenmunson: I'm giving this vulnerability 4 upvotes as, while the vulnerability wasn't the most interesting, how the team handled it was. The fact that it was tagged as related to a bunch of other crashes even though it had nothing to do with them is pretty shocking, but it also goes to show the value of proper debugging and how easy it can be to get red herrings. It's also interesting that the bounty was $1337 ( elite - "leet" ) - it almost seems like the issue was treated as a bit of a joke to the team.

cves/CVE-2011-0776.yml
```diff
@@ -131,12 +132,14 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer: I did not notice any major events during this time.
```
gwenmunson: A bit of a description of the general state of the team here would be nice, or, if nothing of interest can be found, a description of why things weren't eventful (issue was solved quickly, team was rigorously working on it, etc) would help to paint a clearer picture of how the dev team was getting through this issue at the time. It's completely understandable if any major events can't be found, but it's important to give some details of the timeline going on so that anyone reading this can understand how and why this got solved.

cves/CVE-2011-0776.yml
```diff
@@ -51,11 +51,11 @@ fixes_vcc_instructions: |
...
-- :commit: abf23550d6399dc647c984f9135447f658bbaf35
-  :note: ''
+- commit: abf23550d6399dc647c984f9135447f658bbaf35
+  note: Mac block ability to stat arbitrary files in the Sandbox.
 vccs:
-- :commit:
-  :note: 
+- commit: 1e652d0226fa331b2bd80ab7f94e6c4d043c729b
+  note: Mac Refactor sandbox profiles to use a common base configuration file.
 upvotes_instructions: |
```
gwenmunson: I'm giving this 1 upvote as this seems to be a small issue (no bounty) that got fixed quickly/painlessly (no major events).

cves/CVE-2010-3117.yml
jak5779: I see - is there anywhere I should put such a note as this?

cves/CVE-2010-3117.yml
jak5779: True, though I mostly patterned that answer off of the one in the example CVE (CVE-2011-3092).

cves/CVE-2010-3117.yml
```diff
@@ -32,15 +32,15 @@ description_instructions: |
...
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
   was wrong. Otherwise, leave it blank.
 bounty:
-  date: 
-  amount: 
-  references: []
+  amt: 1337.0
+  announced: 2010-08-06
+  url: bugs.chromium.org/p/chromium/issues/detail?id=50553#c22
```
scioffi: Please revert these field name changes. The fields in this file should NEVER change. Please use the original field names for your values.

cves/CVE-2011-0776.yml
```diff
@@ -32,15 +32,15 @@ description_instructions: |
...
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
   was wrong. Otherwise, leave it blank.
 bounty:
-  date: 
-  amount: 
-  references: []
+  amt:
+  announced:
+  url:
```
scioffi: Please revert these field name changes. The fields in this file should NEVER change. Please use the original field names for your values.

cves/CVE-2010-3117.yml
jak5779: I did - these are the field names used in the example document (https://github.com/VulnerabilityHistoryProject/chromium-vulnerabilities/blob/master/cves/CVE-2011-3092.yml), and I could not get it to compile with the other ones.

jak5779: These are the same field names used in the [example document](https://github.com/VulnerabilityHistoryProject/chromium-vulnerabilities/blob/master/cves/CVE-2011-3092.yml). I was having so many difficulties with the YAML syntax that I matched the known-good example as closely as I could. For instance, how do I actually use the array-syntax "references" field as compared to the more straightforward-seeming "url" field?

cves/CVE-2011-0776.yml
jak5779: Same as above

cves/CVE-2010-3117.yml
scioffi: @jak5779 That is interesting. Other PRs are passing with this format. Try comparing against this PR: https://github.com/VulnerabilityHistoryProject/chromium-vulnerabilities/blob/239dff45f442642ab5fd8164a0f39b9558ec45c2/cves/CVE-2014-1743.yml#L43

jak5779: I see - so the array format doesn't actually use the square brackets, maybe?

Either way, I'll have to try it tomorrow.

Maxzolythus:
I give one upvotes to 2010-3117, because it was interesting that this didn't have a CWE. I think this would be something common enough to have one, and that makes me think what else could not have one.

I give one upvotes to 2011-3092, because it was interesting that someone could just escape the sandbox and mutate local files from such a small addition (now subtraction) from the code.

PR #241, opened 2018-10-29
CVE-2010-0660 and CVE-2016-1631

tilley14:
Updated information inside of CVEs 2010-0660 and 2016-1631.

CVE-2010-0660
Disclosure of sensitive data in the referrer header when redirecting from an HTTPS server to an HTTP server.

CVE-2016-1631
XSS in nested modal message loops that bypasses the Same Origin Policy. 

Note:
I was also assigned CVE-2016-1685, a miscalculation of a certain index which allows out-of-bounds reads in PDFium. The code for fixing this is located in a different repository that is maintained specifically for the PDFium plugin. The "fix" in the chromium repository was only updating the DEPS file.

### DIFF
--- a/cves/CVE-2010-0660.yml
+++ b/cves/CVE-2010-0660.yml
@@ -94,8 +94,8 @@ unit_tested:
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
   answer: |
-    From looking at the fix code it's clear that code was tested and they added
-    a new unit test for validating the fix.
+    From looking at the fix code it is clear that the code was tested and that 
+    there were new unit tests for validating the fix.
   code: true
   fix: true
 discovered:
@@ -114,7 +114,7 @@ discovered:
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
   answer: |
-    Found by manually by examining the redirect header after first visiting a URL 
+    Found by manually examining the referrer header after first visiting a URL 
     secured by HTTPS and then clicking a secure HTTPS link that triggers a redirect 
     to an unsecure HTTP URL.
   date: 2009-12-09
@@ -143,10 +143,10 @@ interesting_commits:
   commits:
   - commit: e600c8212f8922d77815316ff41fd8ce9d95bca7
     note: |
-      They added a method to sanatize the referrer header to remove potential usernames 
+      They added a method to sanitize the referrer header to remove potential usernames 
       and passwords. It is not clear from the issue referenced if this change was for 
       automated testing purposes only or if this was an attempt to mitigate leaking sensitive data
-      in the referrer in the field.
+      in the wild.
   - commit: 6568a9e384e0f92422c68d4f31fb401df4acbaed
     note: |
       They worked on some CSRF issues with following a temporary redirect (307 error) for a moved URL.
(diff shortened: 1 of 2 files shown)


cves/CVE-2010-0660.yml
```diff
@@ -181,14 +207,25 @@ lessons:
...
+    applies:
+    note:
+  cognitive_complexity:
+    applies: true
+    note: |
+      The vulnerability involves dependencies on server implementations 
+      outside the control of the browser's developers. The developers have to 
+      brainstorm the possible usages of the referrer header as well as in what 
+      cases it should and should not be sent with a request. This is further 
+      complicated by the number of methods a request can occur and by the number
```
kdm5498: This sentence reads a little bit confusing. Maybe consider "the number of methods a request can use" or something similar.

cves/CVE-2016-1631.yml
```diff
@@ -53,10 +61,10 @@ fixes_vcc_instructions: |
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: dd77c2a41c72589d929db0592565125ca629fb2c
-  :note: ''
+  :note:
 vccs:
-- :commit:
-  :note: 
+- :commit: cd2af395e8429a30bcee9e7ad7ec9f4b680c924d
+  :note:
```
kdm5498: There should be some basic descriptions for these commits to say what they changed.

cves/CVE-2016-1631.yml
```diff
@@ -198,4 +234,18 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    The vulnerability was essentially a design mistake that gave nested modals
+    the privledge to bypass the Same Origin Policy. It is interesting that the
+    two methods used to mitigate this vulnerability were in the repository for 
+    six years before it was discovered.
```
kdm5498: Why do you think it took so long for this mitigation to be implemented in the production code?

cves/CVE-2016-1631.yml
```diff
@@ -198,4 +234,18 @@ mistakes:
...
+
+    It appears that trust boundaries were drawn, but not around all of the components
+    that may cross it. Proper analysis of the architecture is really the only mitigation
+    against this kind of vulnerability. Since this vulnerability was found by someone
+    outside the Chromium team, it is unclear if this has been done.
+    
+    The lesson to be learned here is that it is not good enough to discover a 
+    possible attack vector and fix it in place and then apply this knowledge to
+    future development. This knowledge needs to be applied to the existing code
+    base as well.
```
Fikayoz: What do you mean when you say "This knowledge needs to be applied to the existing code base as well?" Any examples for how this would be done?

Fikayoz:
No issues to report with either CVE-2010-0660 or CVE-2016-1631. Answers on both were detailed and described in an easy-to-understand fashion. Only had one question regarding the answer to CVE-2016-1631

cves/CVE-2010-0660.yml
tilley14: Thanks, by methods a request can use I was trying to refer to POST, GET, and PUT. I think that this will clarify what I meant.

cves/CVE-2016-1631.yml
tilley14: Will do, I wasn't sure if it was needed if they were the only commits

cves/CVE-2016-1631.yml
tilley14: Worst case, I think there may have been another instance of this vulnerability at the time the fixing code was introduced and that the developers did not look through the repo for other possible instances. But more likely, there was similar functionally implemented by a different developer in another system that may have had more knowledge or awareness of this security issue.

cves/CVE-2016-1631.yml
tilley14: I sort of address this in my response to kdm5498. I think that when a vulnerability is discovered of this sort it warrants an audit of the existing subsystems that may be susceptible to this same issue. This could be done with a fuzzer or through manual inspection of the code and existing designs.

PR #254, opened 2018-10-29
CVE-2013-6658 and CVE-2016-5174

falhenaki:
Completed both of CVE-2013-6658.yml and CVE-2016-5174.yml with manual curation

### DIFF
--- a/cves/CVE-2013-6658.yml
+++ b/cves/CVE-2013-6658.yml
@@ -121,7 +121,7 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer: one interesting commit was found.
+  answer: One interesting commit was found.
   commits:
   - commit: 5a430e63a4c764997a1e6612ed1f723acce94461 
     note: in this commit, the author notes in the commit message that this area where the destroy() method is used, is a very risky area in terms of memory allocation. This indicates that the vulnerability could have been avoided as it has been caught between VCC and fix.
--- a/cves/CVE-2013-6658.yml
+++ b/cves/CVE-2013-6658.yml
@@ -170,7 +170,8 @@ lessons:
     note: 
   distrust_input:
     applies: true
-    note: This vulnerability was caused by trusting javscript code running on a site that may induce multiple causees to re-render certain widget that cause a use-after-free.
+    note: |
+      This vulnerability was caused by trusting javscript code running on a site that may induce multiple causees to re-render certain widget that cause a use-after-free.
   security_by_obscurity:
     applies: 
     note: 
--- a/cves/CVE-2013-6658.yml
+++ b/cves/CVE-2013-6658.yml
@@ -32,7 +32,10 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: The vulnerability occurs when Javascript executes multiple triggers for a layout refresh (an attempt to re-render the page at hand). It intermittently fails to handle those multiple triggers and causes a "use-after-free" vulnerability as a triggered component attempts to use another one that has been destroyed (memory freed) by another component caused by the Javascript code which causes a crash (DOS).
+description: The vulnerability occurs when Javascript executes multiple triggers for a layout
+refresh (an attempt to re-render the page at hand). It intermittently fails to handle those multiple
+triggers and causes a "use-after-free" vulnerability as a triggered component attempts to use another
+one that has been destroyed (memory freed) by another component caused by the Javascript code which causes a crash (DOS).
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
@@ -80,7 +83,8 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: It does not appear that unit tests have covered this chunk of code and it appears that unit tests have not been implemented after the fix.
+  answer: It does not appear that unit tests have covered this chunk of code and it
+  appears that unit tests have not been implemented after the fix.
   code: false
   fix: false
 discovered:
@@ -98,7 +102,8 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: The issue was discovered by running a Javascript file that had a number of functions which caused a re-rendering of the browser's layout.
+  answer: The issue was discovered by running a Javascript file that had a number of functions which caused
+  a re-rendering of the browser's layout.
   date: 2013-11-23
   automated: false
   google: false
@@ -124,7 +129,9 @@ interesting_commits:
   answer: One interesting commit was found.
   commits:
   - commit: 5a430e63a4c764997a1e6612ed1f723acce94461 
-    note: In this commit, the author notes in the commit message that this area where the destroy() method is used, is a very risky area in terms of memory allocation. This indicates that the vulnerability could have been avoided as it has been caught between VCC and fix.
+    note: In this commit, the author notes in the commit message that this area where the destroy()
+    method is used, is a very risky area in terms of memory allocation. This indicates that the vulnerability
+    could have been avoided as it has been caught between VCC and fix.
   - commit: 
     note: 
 major_events:
@@ -135,7 +142,8 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: I am not sure if this is a major event but risk was identified. this vulnerability risk was actually noted prior to the original bug reporting during the process of a review. However the other did not mitigate the risk.
+  answer: I am not sure if this is a major event but risk was identified. this vulnerability risk was actually
+  noted prior to the original bug reporting during the process of a review. However the other did not mitigate the risk.
   events:
   - name: Risk Discovery
     date: 2011-07-08
@@ -170,7 +178,8 @@ lessons:
     note: 
   distrust_input:
     applies: true
-    note: This vulnerability was caused by trusting javscript code running on a site that may induce multiple causees to re-render certain widget that cause a use-after-free.
+    note: This vulnerability was caused by trusting javscript code running on a site that may induce multiple
+    causees to re-render certain widget that cause a use-after-free.
   security_by_obscurity:
     applies: 
     note: 
@@ -201,5 +210,9 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: I believe that this vulenrability could have been avoided if mitigations against use-after-free were followed. It seems that between the VCC and fix, it was noted by a review that this area in particular was a high risk area that may induce such a vulnerability. However, no action was taken to mitigate the issue. the vulnerability originated due to a lack of risk mitigation in this high risk area as well as no proper unit tests in place despite this being a system critical component.
+  answer: I believe that this vulenrability could have been avoided if mitigations against use-after-free
+  were followed. It seems that between the VCC and fix, it was noted by a review that this area in particular
+  was a high risk area that may induce such a vulnerability. However, no action was taken to mitigate the issue.
+  the vulnerability originated due to a lack of risk mitigation in this high risk area as well as no proper unit
+  tests in place despite this being a system critical component.
 
--- a/cves/CVE-2013-6658.yml
+++ b/cves/CVE-2013-6658.yml
@@ -4,7 +4,7 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: CWE-399
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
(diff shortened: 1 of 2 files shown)


cves/CVE-2013-6658.yml
```diff
@@ -32,7 +32,7 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: The vulnerability occurs when javascript executes multiple triggers for a layout refresh (an attempt to re-render the page at hand). It intermittently fails to handle those multiple triggers and causes a "use-after-free" vulnerability as a triggered component attempts to use another one that has been destroyed (memory freed) by another component caused by the javascript code.
```
tanner-g:
```suggestion
description: The vulnerability occurs when Javascript executes multiple triggers for a layout refresh (an attempt to re-render the page at hand). It intermittently fails to handle those multiple triggers and causes a "use-after-free" vulnerability as a triggered component attempts to use another one that has been destroyed (memory freed) by another component caused by the javascript code.
```

tanner-g: great description!

cves/CVE-2013-6658.yml
```diff
@@ -80,9 +80,9 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: It does not seem that unit tests have covered this chunk of code and no unit tests have been implemented after the fix was made.
```
tanner-g:
```suggestion
  answer: It does not appear that unit tests have covered this chunk of code and it appears that unit tests have not been implemented after the fix.
```

cves/CVE-2013-6658.yml
```diff
@@ -121,10 +121,10 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer:
+  answer: one interesting commit was found.
```
tanner-g: You need to "demonstrate that you completed this section by explaining what happened between the VCCs and the fix."

cves/CVE-2013-6658.yml
```diff
@@ -121,10 +121,10 @@ interesting_commits:
...
   commits:
-  - commit: 
-    note: 
+  - commit: 5a430e63a4c764997a1e6612ed1f723acce94461 
+    note: in this commit, the author notes in the commit message that this area where the destroy() method is used, is a very risky area in terms of memory allocation. This indicates that the vulnerability could have been avoided as it has been caught between VCC and fix.
```
tanner-g:
```suggestion
    note: In this commit, the author notes in the commit message that this area where the destroy() method is used, is a very risky area in terms of memory allocation. This indicates that the vulnerability could have been avoided as it has been caught between VCC and fix.
```

Make sure to include some of this information in the answer tag above.

cves/CVE-2013-6658.yml
```diff
@@ -201,4 +201,4 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: I believe that this vulenrability could have been avoided if mitigations against use-after-free were followed. It seems that between the VCC and fix, it was noted by a review that this area in particular was a high risk area that may induce such a vulnerability. However, no action was taken to mitigate the issue. the vulnerability originated due to a lack of risk mitigation in this high risk area as well as no proper unit tests in place despite this being a system critical component.
```
TylerLima: Is it possible to still keep the functionality the site still has and still mitigate the risk by "use-after-free" basic coding techniques in a concise change? I think it should be fairly straightforward, but I am guessing there is a reason they haven't done so. Can you think of any reason why?

cves/CVE-2016-5174.yml
```diff
@@ -153,36 +153,36 @@ lessons:
...
-    applies: 
-    note: 
+    applies: true
+    note: Forcing the user to toggle "allow" for every popup may not be the most convenient for the user but it is the most secure option
   yagni:
-    applies: 
-    note: 
-  complex_inputs:
+    applies: true
+    note: This principle has might have led to this vulnerability, it is most likely that the developer of the related functionality either did not consider the flow that would lead to DOS or simply attributed that particular use case to "YAGNI".
```
TylerLima: This principle might has led to the vulnerability and it is most likely that the developer of the related functionality either did not consider the flow that would lead to a DOS or the lack of necessity for fixing this particular use case was deemed appropriate.

cves/CVE-2016-5174.yml
```diff
@@ -197,4 +197,4 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: The origin of this vulnerability is the lack of input validation from the remote target site. In certain conditions, the remote site was able to push popups on to the user's browser, and have them open without asking for user's consent. This, lack of input was due to improper handling of a particular use case in that Flash objects were not filtered for popups and therefore they opened forcibly. By doing this, the user is let vulernable to DOS attack (by spamming the host browser) and redirects to malicious sites. This is likely due to missing security requirements of "Flash" popup handling or the following of yagni principle. The fix, follows the mitigations by CWE. However, I noticed that the fix is only partial. It has resolved the issue of the fullscreen being stuck but popups remain unhandled, which leads me to believe that proper mitigations were not followed and the fix is not proper.
```
TylerLima: Very well explained analysis of the vulnerability and lack of follow through. You said the last commit debated whether or not this issue was "resolved or not" but would you deem their analysis of this CVE as incomplete or perhaps it should have been labeled as a new CVE separate from the initial fullscreen issue?

cves/CVE-2013-6658.yml
```diff
@@ -54,12 +54,12 @@ fixes_vcc_instructions: |
...
-  :note: ''
+  :note: provides comments on the fix code how it works as well as the fix by hardening the vulnerable fuctionality
 - :commit: f93f09e95fad7fb4bdb54035159183104183ac6c
-  :note: ''
+  :note: stop the relayout that caused the vulnerability
 vccs:
-- :commit:
-  :note: 
+- :commit: 5a430e63a4c764997a1e6612ed1f723acce94461
+  :note: this commit introduced the destroy mechanism which destroys a component that may be used by another.
```
scioffi: Please make sure to use proper capitalization and punctuation.

cves/CVE-2013-6658.yml
```diff
@@ -98,20 +98,20 @@ discovered:
...
 subsystem:
   question: |
     What subsystems was the mistake in?
 
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
-  answer: 
-  name: 
+  answer: the mitsake is in the renderer subsystem
```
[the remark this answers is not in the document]
falhenaki: Will do.

cves/CVE-2013-6658.yml
falhenaki: Thanks good catch

cves/CVE-2013-6658.yml
falhenaki: From the conversation I've gathered It seems to have just been too much overhead. The author acknowledged it being a high risk area, he did know/say that it had that vulnerability. 
Quoting the author "RenderWidget::destroy() copied code from various other
        destroy() methods, which made code maintenance in this
        area very risky."

cves/CVE-2016-5174.yml
falhenaki: I agree with your suggestion and have made the change accordingly

cves/CVE-2016-5174.yml
falhenaki: What they did is they separated the vulnerability into two different issues and fixed one of them. The other one (that has not yet been fixed) from my research, is raising a debate on whether it is worth their time considering it is a Flash only issue.

cves/CVE-2013-6658.yml
falhenaki: Thanks, good catch, I believe i took care of all those now.

cves/CVE-2013-6658.yml
```diff
@@ -32,7 +32,7 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: The vulnerability occurs when Javascript executes multiple triggers for a layout refresh (an attempt to re-render the page at hand). It intermittently fails to handle those multiple triggers and causes a "use-after-free" vulnerability as a triggered component attempts to use another one that has been destroyed (memory freed) by another component caused by the javascript code.
```
jahn621: I would also recommend dumbing down the explanation a little more to specifically define what exactly the vulnerability messes with. I would recommend thinking about STRIDE and CIA in regards what the vulnerability may have the potential of ruining.

cves/CVE-2016-5174.yml
```diff
@@ -32,7 +32,7 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: On a browser window, toggle requests go unprocessed which allows attackers (remote sites) to create unsuppressed popups causing a DOS (Denial of Service) by spamming the browser and causing it to crash. Fullscreen mode would also get stuck and you are unable to leave it.
```
jahn621: Good description straight to the point as well as providing a proper example in a real world situation of how people could reproduce this vulnerability.

cves/CVE-2016-5174.yml
```diff
@@ -52,10 +52,10 @@ fixes_vcc_instructions: |
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: 104e5ad762f4d2fcf04a33ff8471d77d7b8678bb
-  :note: ''
+  :note: add code to handle fullscreen stuck issue.
```
jahn621: you mention later down the line that the pop up blocking hadn't been fixed. Is it possible to actually find the commit where that was fixed rather than the full screen fix?

cves/CVE-2016-5174.yml
```diff
@@ -131,10 +131,10 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer: The team did not prioritize stopping popups but ensured that fullscreen mode would not get stuck. a later ticket was opened to block popups (https://bugs.chromium.org/p/chromium/issues/detail?id=626749) but it is still pending and the latest comment has suggested a "WontFix" because the issue only occurs in Flash, which is rarely used.
```
jahn621: Actually i guess this makes sense in regards and answered my previous comment. Either way, that's a pretty good explanation, considering that flash is also dying in less than 2 years.

cves/CVE-2016-5174.yml
```diff
@@ -197,4 +197,5 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
```
jahn621: I really liked this answer. It provides insight into the shortcomings of using Flash and how that could potentially lead to further problems down the line. I do enjoy the CWE connection to and the explanation about the partial fix. For what its worth, knowing that a team jumped ship knowing that one of their tools will soon be obsolete is an interesting thing to consider.  It would be more interesting if they went in the other direction.

jahn621: 4 updoots

cves/CVE-2013-6658.yml
```diff
@@ -201,4 +201,5 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: I believe that this vulenrability could have been avoided if mitigations against use-after-free were followed. It seems that between the VCC and fix, it was noted by a review that this area in particular was a high risk area that may induce such a vulnerability. However, no action was taken to mitigate the issue. the vulnerability originated due to a lack of risk mitigation in this high risk area as well as no proper unit tests in place despite this being a system critical component.
```
jahn621: How would you have approached the risk mitigation in this situation? 

Also 3 Upvotes.

falhenaki: It's very easy to suggest mitigations when you are not familiar with the code base. But, considering that the original author who had familiarity with the code did not fix it, I would assume it is very difficult to mitigate and therefore i do not have a mitigation for this (other than suggesting very careful refactoring).

cves/CVE-2013-6658.yml
```diff
@@ -32,7 +32,7 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: The vulnerability occurs when Javascript executes multiple triggers for a layout refresh (an attempt to re-render the page at hand). It intermittently fails to handle those multiple triggers and causes a "use-after-free" vulnerability as a triggered component attempts to use another one that has been destroyed (memory freed) by another component caused by the Javascript code which causes a crash (DOS).
```
dnp1204: I think you should make this a multiline string, it will make it easier to read

cves/CVE-2013-6658.yml
```diff
@@ -121,10 +121,10 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer:
+  answer: One interesting commit was found.
   commits:
-  - commit: 
-    note: 
+  - commit: 5a430e63a4c764997a1e6612ed1f723acce94461 
+    note: In this commit, the author notes in the commit message that this area where the destroy() method is used, is a very risky area in terms of memory allocation. This indicates that the vulnerability could have been avoided as it has been caught between VCC and fix.
```
dnp1204: Same here I think you should make this a multiline string, it will make it easier to read

PR #257, opened 2018-11-01
Upvotes and a nickname

andymeneely:


mesternefeld:
I like this nickname a lot!

PR #261, opened 2019-01-10
Forked/parameterization

KyleGancarz:
Parameterized start/end dates for git log as well as the git repository location using available environment variables.


andymeneely:
Thanks!

PR #271, opened 2019-10-06
Update CVE-2016-1645.yml - Add a nickname

PuZZleDucK:
Add a nickname to address issue #585 [Study and nickname a vulnerability!].
Also added a commit message I was able to find.


andymeneely:
Thanks! Commit messages get pulled in automatically anyway, but this is fine.

PR #273, opened 2019-10-09
updates CVE-2010-1505 nickname

torianne02:
adds nickname "Off to the Races"
play on words for the race condition getting beat during NTP


andymeneely:
Awesome! Thanks.

PR #275, opened 2019-10-18
Nickname CVE-2010-0644 SOCKS To Be You

mstruve:
Thought the play on "Sucks to be you" was appropriate for this vuln's nickname 😊


andymeneely:
I like it! Thank you very much.

PR #285, opened 2019-10-24
CVE-2016-5173

Jeremy523:
Completed round 1 research on CVE-2016-5173

### DIFF
--- a/cves/CVE-2016-5173.yml
+++ b/cves/CVE-2016-5173.yml
@@ -49,7 +49,7 @@ description: |
   loading in additional Chromium extensions resources allowing them to do things
   such as modifying what happens when objects are garbage-collected, executing
   native functions that typically require user interaction without their
-  interaction, and otherwise call native functions with attacker-controlled data.
+  interaction, and otherwise calling native functions with attacker-controlled data.
 
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
--- a/cves/CVE-2016-5173.yml
+++ b/cves/CVE-2016-5173.yml
@@ -1,24 +1,28 @@
 ---
 CVE: CVE-2016-5173
+
 CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: 284
+
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
+
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
   source for this is Chrome's Stable Release Channel
   (https://chromereleases.googleblog.com/).
   Please enter your date in YYYY-MM-DD format.
-announced: 2016-09-25 16:59:05.807000000 -04:00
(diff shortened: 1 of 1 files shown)


cves/CVE-2016-5173.yml
```diff
@@ -32,38 +36,62 @@ description_instructions: |
...
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
   For the second round of reviewing, you will be giving a certain amount of
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes:
+
```
cxs7700: You can add 10 upvotes here.

cves/CVE-2016-5173.yml
```diff
@@ -197,4 +255,16 @@ mistakes:
...
+    prototype and so the vulnerability wasn't obvious. The main concern comes
+    from the fact that this code is in Chromium's extensions subsystem and has
+    access to many internal never-to-be-exposed resources and native functions.
+    JavaScript at its core is notoriously lax and doesn't do any kind of access
+    control by default on modifying object prototypes. As a result, an attacker
+    in this scenario is able to simply modify the base Object.prototype to
+    add their own functionality and carry out their exploit. The main takeaway
+    from all of this is to expect attackers to abuse the tools you give them.
+    Accounting for this (and understanding how JavaScript works at its core) is
+    key to preventing something like this from happening again in the future.
```
cxs7700: You started off with "The primary mistake", which made it seem like there would be a secondary mistake or more. Earlier in your responses you mentioned that the fix involved a grand total of 50 files and each file was changed the exact same way. Are there any other kinds of mistakes that can be derived from these files?

cves/CVE-2016-5173.yml
```diff
@@ -153,38 +202,47 @@ lessons:
...
+    note:
   secure_by_default:
-    applies: 
-    note: 
+    applies: true
+    note: |
+      The vulnerability centers around the user (or attacker, in this case)
+      misusing components of Chromium's extensions subsystem. In this instance,
+      they did not account for (or prevent) the user modifying the Object.prototype
+      and overriding Chromium's own functionality.
```
cxs7700: Can you further clarify misusing components versus abusing components, possibly by providing a misuse/abuse case for both user and attacker like we did in class?

cves/CVE-2016-5173.yml
```diff
@@ -153,38 +202,47 @@ lessons:
...
   yagni:
-    applies: 
-    note: 
+    applies: true
+    note: |
+      While the developers initially "kept it simple" in their implementation of
+      simply setting Object.prototype bindings directly, this inadvertently 
+      provided "too much functionality" that an attacker can take advantage of.
+      The fix ultimately involved restricting this functionality.
```
cxs7700: For this response and a couple of your other responses, you tend to use parentheses and quotes unnecessarily. You can get the same point across to the reader with or without them.

cves/CVE-2016-5173.yml
Jeremy523: Sounds good

cves/CVE-2016-5173.yml
Jeremy523: Fair points on my wording, I'll improve that. As for all the same changes, yes, I'll write about how they improved the design (and learned from it) to prevent something like this from happening again.

cves/CVE-2016-5173.yml
Jeremy523: I'll clarify a little bit with a simple example of each, but I feel giving a full misuse/abuse case might be going a bit out of scope for this section.

cves/CVE-2016-5173.yml
Jeremy523: Fair point! I can reword a few things to avoid them.

PR #287, opened 2019-10-24
CVE 2015-1282

exl5325:
- Fixed syntax issues in YAML
- Set curated to true
- Added clarifying details
- Added missing details

### DIFF
--- a/cves/CVE-2015-1282.yml
+++ b/cves/CVE-2015-1282.yml
@@ -116,7 +116,12 @@ discovered:
   answer: |
     This vulnerability was discovered by a user. That same user also submitted the fix
     for the vulnerability. No information was available on the user's profile. The user
-    did not provide a name. The code review for the bug was private.
+    did not provide a name. 
+    
+    The user had an example PDF which exploited the vulnerability, and tested the fix by 
+    using the same PDF.
+    
+    The code review for the bug was private.
   date: 2015-05-14
   automated: false
   google: false
--- a/cves/CVE-2015-1282.yml
+++ b/cves/CVE-2015-1282.yml
@@ -33,12 +33,22 @@ description_instructions: |
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
 description: |
-  Crafted javascript in a PDF file could be used to crash chrome or potentially
-  lead to arbitrary code execution. This vulnerability is caused by javascript causing
-  the PDFium (PDF) engine to enter a method, multiple times by causing document level and
-  field level java script to interact using the field.delay property. This would free
-  an already freed pointer. This vulnerability would trigger Chromium's AddressSanitizer, a
-  memory checking library, which would crash Chromium so further damage could be prevented.
+  In Chromium, PDF's are handled by an open sourced library called PDFium. PDFium
+  was developed along with Chromium, however the library remained closed source until 2014.
+  This vulnerability was discovered in 2015.
+
+  Adobe (the creators of the PDF standards) includes a Javascript API that can allow
+  PDF's to embed and execute JavaScript code in order to enhance a PDF's functionality. For ex,
+  PDF's that include buttons are typically written with Javascript code. 
+
+  This vulnerability allows a "crafted" PDF (term taken from Chromium bug report) that 
+  contains specific Javascript code to crash the browser. Javascript code can cause
+  the delay() and doFieldDelay() functions in Document.cpp to call free() twice. This vulnerability 
+  is caused by javascript causing the PDFium (PDF) engine to enter a method, multiple times by 
+  causing document level and field level JavaScript to interact using the field.delay property.
+  This would free a previously freed pointer. This results in a denial of service attack, 
+  since Chromium defaults to crashing if it detects a use-after-free bug. 
+
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
--- a/cves/CVE-2015-1282.yml
+++ b/cves/CVE-2015-1282.yml
@@ -192,10 +192,10 @@ lessons:
   least_privilege:
     applies: true
     note: |
-      Principal of least privilege does not apply directly to this vulnerability. However, allowing
-      Javascript in PDF's is an inherent vulnerability in of it itself. One good coding principal is
-      to sandbox any executing Javascript as much as possible, due to the high probability that a user
-      will open a malicious PDF.
+      Reducing the capabilites of the JavaScript API in PDF's as much as possible is a 
+      mitigation that should be considered. This could fall in line with YAGNI - 
+      Adobe (the makers of the PDF standard) should look at how much of the JS API is 
+      actually needed for a good average user experience.
   frameworks_are_optional:
     applies:
     note:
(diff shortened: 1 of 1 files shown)


cves/CVE-2015-1282.yml
```diff
@@ -179,8 +190,12 @@ lessons:
     note: Chrome's AddressSanitizer caught a bad memory operation and forced the application
       to crash before any damage could be done.
   least_privilege:
-    applies:
-    note:
+    applies: true
+    note: |
+      Principal of least privilege does not apply directly to this vulnerability. However, allowing
```
CatherineOsadciw: I would classify this as distrust input instead -- good observation though!

exl5325: > I would classify this as distrust input instead

I agree. Thanks for the suggestion. I moved it to distrust input, and also added some new thoughts in `least privilege` and `YAGNI`.

cves/CVE-2015-1282.yml
```diff
@@ -116,11 +116,16 @@ discovered:
   answer: |
     This vulnerability was discovered by a user. That same user also submitted the fix
     for the vulnerability. No information was available on the user's profile. The user
-    did not provide a name. The code review for the bug was private.
-  date: 2015/05/14
-  automated:
+    did not provide a name. 
+    
+    The user had an example PDF which exploited the vulnerability, and tested the fix by
```
Rxy8841: Could you show a snippet of the example, it would be beneficial to see an actual example.

cves/CVE-2015-1282.yml
```diff
@@ -139,7 +144,13 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer:
+  answer: |
+    It appears that this vulnrability had existed in PDFium within Chromium for a significant period of
```
Rxy8841: That is crazy 13 months! You would think such a highly used browser would have found that problem quicker.

cves/CVE-2015-1282.yml
```diff
@@ -179,8 +190,12 @@ lessons:
     note: Chrome's AddressSanitizer caught a bad memory operation and forced the application
```
Rxy8841: This is a perfect example of using defense in depth.

cves/CVE-2015-1282.yml
```diff
@@ -205,11 +226,19 @@ lessons:
...
-    note:
+    applies: true
+    note: See comments in least privilege section. 
   complex_inputs:
     applies:
     note:
+  fail_securely:
+    applies: true
+    note: |
+      The interesting tidbit is that this vulnerability was actually exploiting code designed to
```
Rxy8841: This is a good example of how the exploit could be used.

cves/CVE-2015-1282.yml
```diff
@@ -223,10 +243,16 @@ mistakes:
...
-    This one is tough to answer because the history of this bug is older than the available
-    code base and the code review was made private by the bug finder/fixer. I have almost
-    no insight to what was going on.
+    The mistake was a design mistake and a coding mistake. One thing the developers did well was
+    that they followed the fail securely principal. Freeing a pointer twice could cause undeterministic
+    outcomes - as a result it is best to terminate when such an error occurs. However, most people
+    do not consider a vulnerability fix or good coding practice to also something that can be exploited
+    for malicious purposes. That was the case in this scenario. Knowing a foolproof mechanism to
+    crash the browser lead to malicious PDF's that could simply invoke that failsafe over and over - 
+    writing a PDF that forced the browser to free a pointer twice, intentionally crashing it.
```
adw4236: Your absolutely right, crashing the program when a security vulnerability occurs does not fix the security vulnerability, it just changes the category.  It's a bandaid on a larger problem.

adw4236: Is crashing really the best solution here? Maybe a better option would be to have a custom free that sets the freed pointer to null so that it can't be freed again like we discussed in class.  Would this work in a large system like this? Maybe it would add too much complexity and should only be considered for this case.

cves/CVE-2015-1282.yml
```diff
@@ -223,10 +243,16 @@ mistakes:
...
+
+    One potential solution that could be considered is to forbid Javascript in PDF's entirely - JS
+    in PDF's is an enormous security risk by itself that there should be debate whether or not the benefit
+    of having a Javascript API is worth the security headaches.
```
adw4236: Forbidding JavaScript entirely seems a bit drastic, if JavaScript is that exploitable, wouldn't it have faced the same issues as Flash and died on the web at this point? I don't think JavaScript is inherently a bad thing, but having a much more restricted API would help keep unexpected interactions to a minimum.

exl5325: > It's odd that most of this file was already filled out, maybe you could go through and ensure everything is clear that was preset? E.g. "this would free a freed pointer" could be worded a bit better.
> 
> Someone must have curated this prior, I wonder if they just forgot to mark it as curated?

Yeah, it was curated by a prior student. I mentioned it to the professor and he said to stick with it and add on to it. I'll take a look at the wording of some of the older comments

exl5325: > if JavaScript is that exploitable, wouldn't it have faced the same issues as Flash and died on the web at this point

I don't think this is a fair comparison - Javascript is a language while Flash is a platform. I also disagree on the wording, but maybe I wasn't clear myself. Javascript itself isn't inherently exploitable, but Javascript code can be very easily abused to be exploitable. Obviously, the same can be said about any language, but I mention this because I think it is worth debating whether the benefits of Javascript in PDF outweigh its cons. I'm not too versed on the Acrobat JS API (I didn't even know JS was in PDF's until today), but I thought it would be an interesting debate

adw4236: That's a fair point, but I think there are other solutions to be discussed as well that avoid the need to completely remove JS from PDFs.

cves/CVE-2015-1282.yml
exl5325: Example PDF is attached in the bug report on the Chromium site. They are just blank PDF's with Javascript embedded in them, so there isn't much to see.

cves/CVE-2015-1282.yml
exl5325: I added a potential mitigation (restricting the JS API) along with the potential drawbacks of such a solution

cves/CVE-2015-1282.yml
exl5325: I think a custom `free()` is just a whack a mole solution because doing so will simply open up a serious potential bug that would affect the rest of the system - if the custom `free`'s checks are written incorrectly, it could prevent memory from being freed at all and cause memory leaks everywhere. I think the best solution is to follow the language's coding specification probably and ensure that `free()` can never be called twice

cves/CVE-2015-1282.yml
```diff
@@ -139,7 +139,13 @@ interesting_commits:
...
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer:
+  answer: |
+    It appears that this vulnrability had existed in PDFium within Chromium for a significant period of
+    time. PDFium started off as a closed source project before being open-sourced in June 2014. This vulnerability
+    was discovered approximately 13 months later in July 2015. According to git blame, the code in question was written
+    before PDFium was open-sourced, making it impossible to trace the origins of the vulnerability. There were no changes
+    to the lines of code in question during the 13 months between open source and discovery. Therefore, there are no remarkable
+    commits to note.
```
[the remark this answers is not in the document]
exl5325: I dug around and didn't find anything of note - big problem was the one I mentioned in the YAML - most of this specific subsystem was closed source when this vulnerability was introduced

PR #294, opened 2019-10-25
CVE-2008-6998 updated all the fields

Rxy8841:
CVE-2008-6998 updated all the fields

### DIFF
--- a/cves/CVE-2008-6998.yml
+++ b/cves/CVE-2008-6998.yml
@@ -65,8 +65,7 @@ fixes:
 vccs:
 - :commit: 09911bf300f1a419907a9412154760efd0b7abc3
   :note: | 
-   This is the initial commit for chrome, they never thought to check the input
-   of the user.
+   Add Chrome to the repository
 upvotes_instructions: |
   For the first round, ignore this upvotes number.
 
@@ -160,10 +159,12 @@ major_events:
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
   answer: |
-    No the problem was there initially and was never found or fixed up to the point of the commit. 
-    They were not careful in the code or the tests they created and missed the corner case. 
+    No the problem was there initially, and was never found or fixed up to the point of the commit. 
+    Some major events I found was that a lot of chromes problems were found by automated testing.
+    Which is weird because this vunerability was not caught by the automated tests. Most of the commits
+    were actually for fixing or adding automated tests.
   events:
-  - name: 
+  - name:
     date: 
   - name: 
     date: 
@@ -199,7 +200,9 @@ lessons:
     applies: true
     note: |
         As a programmer they should always becareful with user inputs and never trust a user 
-        to use the system correctly. They should put more checks and security around it.
+        to use the system correctly. In this case the url previewer broke and cause a system 
+        shutdown which allowed the user to run arbitrary code. By checking the input for 
+        malformed text and length, the devleopers will be able to solve and prevent this problem.
   security_by_obscurity:
     applies: 
     note: 
@@ -218,7 +221,7 @@ lessons:
   complex_inputs:
     applies: true
     note: |
-        The input was not expected, and some users could use the fact and execute arbitrary code
+        The URL was not expected, and some users could use the fact and execute arbitrary code
         with the vulnerability.
 mistakes:
   question: |
--- a/cves/CVE-2008-6998.yml
+++ b/cves/CVE-2008-6998.yml
@@ -33,10 +33,10 @@ description_instructions: |
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
 description: |
-  Google chrome had a problem when a user typed in a very long and malformed url. 
-  The backend array that held the url typed in would overflow and shut the system
-  down. It could also lead to users being able to run arbutrary code after the 
-  system crashes. 
+  Google chrome had a problem when a user typed in more than 128 chars in the url, 
+  then hovered over the url. The url preview should show up but the backend array 
+  that held the url typed in would overflow and shut the system down. It could also 
+  lead to users being able to run arbitrary code after the system crashes. 
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
@@ -57,7 +57,9 @@ fixes_vcc_instructions: |
 fixes:
 - commit: 29d1eb815f296355e63a555729abc17b896d220a
   note: | 
-   They added a check on the input to make sure the length of the input is not longer then the size of there buffer. 
+   The developers added a check for the URL preview functionality to make sure it doesnt try
+   and display a large URL. If the URL is longer than 128 chars then it would only
+   show up to 128 chars. 
    
 vccs:
 - :commit: 09911bf300f1a419907a9412154760efd0b7abc3
@@ -71,7 +73,7 @@ upvotes_instructions: |
   upvotes to each vulnerability you see. Your peers will tell you how
   interesting they think this vulnerability is, and you'll add that to the
   upvotes score on your branch.
-upvotes: 
+upvotes: 2
 unit_tested:
   question: |
     Were automated unit tests involved in this vulnerability?
@@ -85,7 +87,9 @@ unit_tested:
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
   answer: |
-    No there were no unit tests for this problem. They had this problem since they commited the code. Yes haveing the automated testes catch this corner case will prevent this problem from happening again. 
+    No, there were no unit tests for this problem. They had this problem since they 
+    commited the code. By adding this corner case into automated tests it will prevent 
+    this problem from happening again. 
   code: false
   fix: true
 discovered:
@@ -104,7 +108,10 @@ discovered:
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
   answer: |
-    I could not find anything about who or how they found the problem. But it was probably from an outside source. But I went onto the CVE and looked though all the BID and exploit urls. I sereched through all of them but they only provided what happened and the type of vulnerability it was. 
+    I could not find anything about who or how they found the problem. But it was probably from 
+    an outside source. I went onto the CVE and looked though all the BID and exploit urls. 
+    I sereched through all of them but they only provided what happened and the type of vulnerability 
+    it was. 
     
   date: 
   automated: 
@@ -129,7 +136,13 @@ interesting_commits:
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
   answer: |
-    There were actually a few buffer overflow vunerability problem in that one days worth of commits. It seems like the code itself was missing a lot of checks and was allowing users to input anything they wanted to. This is falling into the problem of secure by default, and security by obscurity where they thought the system was secure and users would not do anything weird or unecessary. That was from one of the bug report readings on the CVE. But the commits that occured on that day included a chrome string update, they disabled tests for a bug, a bunch of broken builds and roll backs. 
+    There were actually a few buffer overflow vunerability problem in that one days worth of 
+    commits. It seems like the code itself was missing a lot of checks and was allowing any 
+    user inputs and placing them right into their arrays. This is falling into the problem of 
+    security by obscurity where they thought the system was secure and users would not do 
+    anything weird or unecessary. These events was from one of the bug report readings on the CVE. 
+    The commits that occured on that day included a chrome string update, they disabled tests for 
+    a bug, a bunch of broken builds and roll backs. 
   commits:
   - commit: 42a9ab0deed87ec463c7fd49e7d206380c9ee0cb
     note: |
@@ -146,7 +159,8 @@ major_events:
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
   answer: |
-    No the problem was there initially and was never found or fixed up to the point of the commit. They were not careful in the code or the tests they created and missed the corner case of the system. 
+    No the problem was there initially and was never found or fixed up to the point of the commit. 
+    They were not careful in the code or the tests they created and missed the corner case. 
   events:
   - name: 
     date: 
@@ -183,7 +197,8 @@ lessons:
   distrust_input:
     applies: true
     note: |
-        As a programmer they should always becareful with user inputs and never trust a user to use the system correctly. They should put more checks and security around it.
+        As a programmer they should always becareful with user inputs and never trust a user 
+        to use the system correctly. They should put more checks and security around it.
   security_by_obscurity:
     applies: 
     note: 
@@ -196,7 +211,8 @@ lessons:
   secure_by_default:
     applies: true
     note: |
-        They should have put a check into the user input, they probably pushed that idea to the side and never thought a user could break the url input. 
+        They should have put a check into the user input, they probably pushed that idea to 
+        the side and never thought a user could break the url input. 
   yagni:
     applies: 
     note: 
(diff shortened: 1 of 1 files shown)


cves/CVE-2008-6998.yml
```diff
@@ -32,7 +32,11 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
+  Google chrome had a problem when a user typed in a very long and malformed url. 
+  The backend array that held the url typed in would overflow and shut the system
+  down. It could also lead to users being able to run arbutrary code after the
```
exl5325: nit: spelling (arbitrary)

cves/CVE-2008-6998.yml
```diff
@@ -32,7 +32,11 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: |
```
exl5325: I think some more detail in the description would be useful. Users type in long and malformed URL's all the time inadvertently - yet that doesn't always cause the browser to crash. What is exactly is a "very long" URL (ie. something with more than x number of characters)? Does the same issue occur if a user types in a super long string that happens to be valid?

cves/CVE-2008-6998.yml
```diff
@@ -51,11 +55,15 @@ fixes_vcc_instructions: |
   Please put the commit hash in "commit" below (see my example in
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
-- :commit: 29d1eb815f296355e63a555729abc17b896d220a
-  :note: ''
+- commit: 29d1eb815f296355e63a555729abc17b896d220a
+  note: | 
+   They added a check on the input to make sure the length of the input is not longer then the size of there buffer.
```
exl5325: nit: "there" -> "their"

In general, I think its good practice to avoid vague pronouns - By "they" I assume you mean "The developers" and "their buffer" I assume refers to the buffer in memory used to hold in the URL. It's better to just explicitly state this.

cves/CVE-2008-6998.yml
```diff
@@ -76,9 +84,10 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: |
+    No there were no unit tests for this problem. They had this problem since they commited the code. Yes haveing the automated testes catch this corner case will prevent this problem from happening again.
```
exl5325: nit (spelling): haveing -> having, testes (haha) -> tests. 

I'd also add some commas after No in the first sentence and Yes to the second

cves/CVE-2008-6998.yml
```diff
@@ -117,12 +128,15 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer:
+  answer: |
+    There were actually a few buffer overflow vunerability problem in that one days worth of commits. It seems like the code itself was missing a lot of checks and was allowing users to input anything they wanted to. This is falling into the problem of secure by default, and security by obscurity where they thought the system was secure and users would not do anything weird or unecessary. That was from one of the bug report readings on the CVE. But the commits that occured on that day included a chrome string update, they disabled tests for a bug, a bunch of broken builds and roll backs.
```
exl5325: nit: vunerability -> vulnerabilities, problem -> problems, days -> day's

Interesting read. It's surprising to know that such a established company would make so many oversights when it came to code quality. I don't see how it relates to the secure by default principle though. Secure by default refers to defaulting config settings to the most secure settings possible - this was more of an error in coding than configuration

cves/CVE-2008-6998.yml
```diff
@@ -165,8 +181,9 @@ lessons:
     applies: 
     note: 
   distrust_input:
-    applies: 
-    note: 
+    applies: true
+    note: |
+        As a programmer they should always becareful with user inputs and never trust a user to use the system correctly. They should put more checks and security around it.
```
exl5325: nit: becareful -> be careful

cves/CVE-2008-6998.yml
```diff
@@ -153,8 +168,9 @@ lessons:
     If you think of another lesson we covered in class that applies here, feel
     free to give it a small name and add one in the same format as these.
```
exl5325: I think the `fail securely` principle would be a good candidate to be added for this vulnerability.

cves/CVE-2008-6998.yml
```diff
@@ -197,4 +217,5 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    This was a design mistake. I think this was a design mistake because the had other overflow problems that followed with the commit of the url overflow problem. So they were missing a crucial check of array size in there code. Which means they were overlooking the problem entirely. Also make sure the user inputs are alwyas checked with white and black lists. This will solve problems before they even come up. Solving the problem as they show up when the system is released is much harder to spot all the errors. Implement the checks and input validations as they were programming.
```
exl5325: Careful of "there" vs "their".

While I can see how this is a design problem, I'd argue that it is more of a coding mistake than a design mistake. You typically don't see nitty gritty implementation details such as checking the length of inputs discussed during the architecture or design phase. Based off what I read, this seems to be more of an oversight from developers during the implementation phase, so I'd counter that it's a coding mistake.

cves/CVE-2008-6998.yml
Rxy8841: Spelling errors will all be fixed!!! <- This will serve for all grammar comments. Sorry there was so much.

cves/CVE-2008-6998.yml
Rxy8841: Great question, I should not have used the word "malformed". If the user typed more than 128 chars into the URL and hovered over the URL to get the URL preview. The backend function for that has an array that overflowed and crashed the system.

cves/CVE-2008-6998.yml
Rxy8841: You are correct, this is not a problem under secure by default. I agree that it is a coding mistake, they should have been more careful with the checks on the size of their arrays.

cves/CVE-2008-6998.yml
Rxy8841: I see where the principle could be applied, but with array size checking there would be no need for the system to fail securely in this vulnerability. Due to the simplicity of the function the vulnerability appeared in, there is no need for another check to make sure it worked.

cves/CVE-2008-6998.yml
Rxy8841: I was also having problems on choosing which one it was. I still think it was a big part of the design phase. So I will update the answer to have both mistakes a part of the entire problem.

cves/CVE-2008-6998.yml
exl5325: > If the user typed more than 128 chars into the URL and hovered over the URL to get the URL preview. The backend function for that has an array that overflowed and crashed the system

This would definitely be great (and interesting) detail to add in!

exl5325:
I'd like to add 2 upvote points for this vulnerability.

cves/CVE-2008-6998.yml
```diff
@@ -76,9 +84,10 @@ unit_tested:
```
exl5325: I'd like to add 2 upvote points for this vulnerability

cves/CVE-2008-6998.yml
adw4236: How would you go about injecting arbitrary code? By having a pointer to a method at the end of your URL?  Does this pose a problem when someone clicks on links or is this just if the user types in the URL manually? If it's the former, that seems like a much bigger issue than the latter.

cves/CVE-2008-6998.yml
```diff
@@ -94,7 +107,12 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
+  answer: |
+    I could not find anything about who or how they found the problem. But it was probably from 
+    an outside source. I went onto the CVE and looked though all the BID and exploit urls. 
+    I sereched through all of them but they only provided what happened and the type of vulnerability 
+    it was.
```
adw4236: I'm not sure if these should be written in the first person? I was under the impression it should be third, but I could be wrong.

cves/CVE-2008-6998.yml
```diff
@@ -106,8 +124,8 @@ subsystem:
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
-  answer: 
-  name: 
+  answer: c++ compiler, the code is under url_elider.
```
adw4236: What does the C++ compiler have to do with the subsystem?

cves/CVE-2008-6998.yml
```diff
@@ -117,12 +135,21 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer:
+  answer: |
+    There were actually a few buffer overflow vunerability problem in that one days worth of 
+    commits. It seems like the code itself was missing a lot of checks and was allowing any 
+    user inputs and placing them right into their arrays. This is falling into the problem of 
+    security by obscurity where they thought the system was secure and users would not do 
+    anything weird or unecessary. These events was from one of the bug report readings on the CVE.
```
adw4236: This doesn't quite seem like security by obscurity, there isn't sensitive resources that they are trying to hide behind something hard to guess, it's more like complex input where they are not limiting the things that can be entered in.

cves/CVE-2008-6998.yml
```diff
@@ -177,14 +209,18 @@ lessons:
     applies: 
     note: 
   secure_by_default:
-    applies: 
-    note: 
+    applies: true
+    note: |
+        They should have put a check into the user input, they probably pushed that idea to 
+        the side and never thought a user could break the url input.
```
adw4236: I'm not sure if secure by default applies here. From my understanding, that principle is about ensuring your configuration is as secure as possible by default and force people to have to think about turning off security settings.

cves/CVE-2008-6998.yml
```diff
@@ -197,4 +233,14 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: |
+    This was bot a design  and coding mistake. I think this was a design mistake because they
```
adw4236: "bot" -> "both"
there are two spaces after design

cves/CVE-2008-6998.yml
```diff
@@ -197,4 +233,14 @@ mistakes:
...
-  answer: 
+  answer: |
+    This was bot a design  and coding mistake. I think this was a design mistake because they 
+    had other overflow problems that followed with the commit of the url overflow problem. So 
+    they were missing a crucial check of array size in their code. Which means they were overlooking 
+    the problem entirely. This could also fall into the category of coding mistakes because nit pick
+    errors don't really belong in the design phase.
+    
+    Some other fixes they should add are making sure the user inputs are always checked with white 
+    and black lists. This will solve problems before they even come up. Solving the problem as they
```
adw4236: Would white/blacklists actually help here? If you are just talking about the size of a string, white-listing or blacklisting certain words or phrases isn't going to help keep the size down.

cves/CVE-2008-6998.yml
adw4236: What language was this written in? I have been assuming C, but I'm not sure if I read that somewhere or just assumed.

cves/CVE-2008-6998.yml
```diff
@@ -197,4 +233,14 @@ mistakes:
...
+    This was bot a design  and coding mistake. I think this was a design mistake because they 
+    had other overflow problems that followed with the commit of the url overflow problem. So 
+    they were missing a crucial check of array size in their code. Which means they were overlooking 
+    the problem entirely. This could also fall into the category of coding mistakes because nit pick
+    errors don't really belong in the design phase.
+    
+    Some other fixes they should add are making sure the user inputs are always checked with white 
+    and black lists. This will solve problems before they even come up. Solving the problem as they 
+    show up when the system is released is much harder to spot all the errors. Implement the checks 
+    and input validations as they were programming.
```
adw4236: How do you think someone should go about ensuring that all fields are of a valid length.  In most high level languages, strings are flexible and you don't have to worry about overflowing, but C is different.  Is there some architectural decision you could make to prevent things like this from happening?

exl5325:
Random tip but I suggest clicking on "resolve conversation" for each issue you've already addressed so that it's easier to keep track of which ones you need to follow up on

cves/CVE-2008-6998.yml
Rxy8841: So in C you can check the sizeof an input with sizeof(). Which is what they do on the URL input, and then they only show a part of the URL in the preview.

cves/CVE-2008-6998.yml
Rxy8841: Sorry this was not specifically for this problem. It was a extra precaution I thought they should take to make sure Users are not trying to break the system. Ex.) User tries to pass sql commands. The URL should not allow it just in case. Sorry for the confusion.

cves/CVE-2008-6998.yml
Rxy8841: fixed

cves/CVE-2008-6998.yml
Rxy8841: That is correct, I thought security by default was saying developers don't implement security measures because they think its secure.

cves/CVE-2008-6998.yml
Rxy8841: When I think about Security By Obscurity, the developers did not protect something because they thought the users would not be able to attack that functionality. But I do see where your coming from that this vulnerability is because they did not catch the problem. But I still believe the principle applies in this case.

cves/CVE-2008-6998.yml
Rxy8841: The code for this subsystem was written in C++. In the example on the history project, it says c compiler because the subsystem was written in that language. That is how I read the example.

cves/CVE-2008-6998.yml
Rxy8841: The question asked "write down where you looked for the answer." Which is why I wrote in first person. I don't know the answer to this as well.

cves/CVE-2008-6998.yml
Rxy8841: So the problem only is occurring in the URL preview when you hover over the URL. The array for the URL previewer was to small compared to the actual URL. 

I will describe the execution of the arbitrary code better! It is written in C++

cves/CVE-2008-6998.yml
```diff
@@ -51,19 +56,25 @@ fixes_vcc_instructions: |
...
+   The developers added a check for the URL preview functionality to make sure it doesnt try
+   and display a large URL. If the URL is longer than 128 chars then it would only
+   show up to 128 chars. 
+   
 vccs:
-- :commit:
-  :note: 
+- :commit: 09911bf300f1a419907a9412154760efd0b7abc3
+  :note: | 
+   This is the initial commit for chrome, they never thought to check the input
```
CatherineOsadciw: Note should be the commit message

cves/CVE-2008-6998.yml
CatherineOsadciw: Remove references to yourself and just put something like "Unknown source"

cves/CVE-2008-6998.yml
```diff
@@ -131,7 +159,9 @@ major_events:
 
     The event doesn't need to be directly related to this vulnerability, rather,
     we want to capture what the development team was dealing with at the time.
-  answer: 
+  answer: |
+    No the problem was there initially and was never found or fixed up to the point of the commit.
```
CatherineOsadciw: Were there any major events that happened between when the VCC was introduced and when it was fixed?

cves/CVE-2008-6998.yml
```diff
@@ -165,8 +196,10 @@ lessons:
     applies: 
     note: 
   distrust_input:
-    applies: 
-    note: 
+    applies: true
+    note: |
+        As a programmer they should always becareful with user inputs and never trust a user
```
CatherineOsadciw: Answer in the context of this specific vulnerability

cves/CVE-2008-6998.yml
```diff
@@ -177,14 +210,16 @@ lessons:
...
+    note:
   yagni:
     applies: 
     note: 
   complex_inputs:
-    applies: 
-    note: 
+    applies: true
+    note: |
+        The input was not expected, and some users could use the fact and execute arbitrary code
```
CatherineOsadciw: What was the input for this vulnerability?

Rxy8841: Sorry the input is the URL.

cves/CVE-2008-6998.yml
Rxy8841: I have added a section tying it back to the vulnerability. The section was to broad and didn't explain more about the vulnerability. Thank You

cves/CVE-2008-6998.yml
Rxy8841: I will update the section with more information.

cves/CVE-2008-6998.yml
Rxy8841: I have gone through and tried to remove all the Self references I could. But I left some where it was necessary.

iclare:
1 upvote - you would think such a frequently used component wouldn't have such an issue

Nitepone:
I will give this 3 upvotes

PR #296, opened 2019-10-25
CVE-2015-1281

Moremothansasuke:
This file was mostly done before it was assigned to me but curated was not set to true.

### DIFF
--- a/cves/CVE-2015-1281.yml
+++ b/cves/CVE-2015-1281.yml
@@ -18,7 +18,7 @@ announced_instructions: |
   source for this is Chrome's Stable Release Channel
   (https://chromereleases.googleblog.com/).
   Please enter your date in YYYY-MM-DD format.
-announced: 2015-07-22 20:59:10.693000000 -04:00
+announced: 2015-07-22
 description_instructions: |
   You can get an initial description from the CVE entry on cve.mitre.org. These
   descriptions are a fine start, but they can be kind of jargony.
@@ -33,14 +33,15 @@ description_instructions: |
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
 description: Some Chromium extensions such as Adblock Pro executed JavaScript that
-  would break CSP (Content Security Policy) checks.  This allows an attacker to load
-  an arbitrary image which can potentially be used to grab user's sensitive information.
+  would break CSP (Content Security Policy) checks. Attackers are then able to load
+  any image that they would like, and possibly use the image's request to access the
+  user's sensitive data.
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
   was wrong. Otherwise, leave it blank.
 bounty:
-  date: 2015-07-21 14:00:00.000000000 -04:00
+  date:
   amount: 1000.0
   references:
   - http://chromereleases.googleblog.com/2015/07/stable-channel-update_21.html
@@ -108,7 +109,7 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: Reported by someone not in Google partial email address of masatoki...@gmail.com.
+  answer: Reported by someone not in Google with a partial email address of masatoki...@gmail.com.
   date: 2015-05-12
   automated:
   google: false
@@ -120,8 +121,8 @@ subsystem:
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
-  answer: Blink->SVG
-  name: Blink->SVG
+  answer: Blink, SVG
+  name: Blink, SVG
 interesting_commits:
   question: |
     Are there any interesting commits between your VCC(s) and fix(es)?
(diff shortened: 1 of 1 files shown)


cves/CVE-2015-1281.yml
```diff
@@ -58,13 +58,13 @@ fixes_vcc_instructions: |
...
+  :note: Keeps track of isolates so that everything can be isolated properly.
 - :commit: 9fb1e9fc1f3829f9597f830e599fc40fadc84db3
-  :note: Reimplement fix this time it passed tests.  Last commit for this fix.
+  :note: Reimplement fix for properly setting ScriptState for image loading.
 - :commit: 7faa3067e4b664eb19a7a682697f5b387a130728
-  :note: 
+  :note: Properly set ScriptState for image loading.
 - :commit: 0f291b9f331a04c55e0d050bfaef20b1c6a6af1c
-  :note: Revert attempt at fix as it was failing tests.
+  :note: Revert 3rd commit in this list (7faa) due to possibly causing leaking issues.
```
CatherineOsadciw: Make sure notes are the commit messages

cves/CVE-2015-1281.yml
```diff
@@ -58,13 +58,13 @@ fixes_vcc_instructions: |
...
 vccs:
 - :commit: 52aa3b1bcf3079fc72a2b24a73dff78017eb5e7a
   :note: Bug fix for issue 200923002.  The issue was fixed by adding a microtask to
```
CatherineOsadciw: Same as above, note should be the commit message

cves/CVE-2015-1281.yml
```diff
@@ -110,17 +110,17 @@ discovered:
...
 subsystem:
   question: |
     What subsystems was the mistake in?
 
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
-  answer: 
+  answer: Blink->SVG
   name: Blink->SVG
```
CatherineOsadciw: Change to "Blink, SVG"

cves/CVE-2015-1281.yml
```diff
@@ -11,7 +11,7 @@ curated_instructions: |
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
```
CatherineOsadciw: Remove timestamp from "announced" below (leave date)

CatherineOsadciw: and bounty date below

cves/CVE-2015-1281.yml
```diff
@@ -58,13 +58,13 @@ fixes_vcc_instructions: |
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
 fixes:
 - :commit: dff368031150a1033a1a3c913f8857679a0279be
-  :note: 
+  :note: Keeps track of isolates so that everything can be isolated properly.
```
CatherineOsadciw: Keeps track of isolates? Not sure what you're referring to here.

cves/CVE-2015-1281.yml
```diff
@@ -171,36 +171,36 @@ lessons:
...
-    applies: 
-    note: 
+    applies:
+    note:
   yagni:
-    applies: 
-    note: 
+    applies:
+    note:
   complex_inputs:
```
kzalbak: Won't let me comment below, but I would consider talking about the mitigation further in your analysis. Talk about what the fix actually accomplished. It also seems really choppy with all the short sentences, consider expanding the sentences for better flow.

cves/CVE-2015-1281.yml
```diff
@@ -110,17 +110,17 @@ discovered:
     leave the entries blank except for "answer". Write down where you looked in "answer".
   answer: Reported by someone not in Google partial email address of masatoki...@gmail.com.
```
bnk5096: Maybe refine this answer a little bit.
Perhaps: "Reported by someone not working at Google. The reporter has a partial email address of  ..."

cves/CVE-2015-1281.yml
```diff
@@ -58,13 +58,13 @@ fixes_vcc_instructions: |
   CVE-2011-3092.yml). Fixes and VCCs follow the same format.
```
bnk5096: I can't comment above, but maybe refine the description. it seems to be a bit high level. How is this dangerous javascript injected? 

If you don't decide to do a refactor here, `user's` should be either `a user's` or `users'`

cves/CVE-2015-1281.yml
Moremothansasuke: An isolate is a chromium specific term that refers to a bunch of data that should be kept isolated and not confused with other data, like how each tab is separate from each other tab. I don't know if chromium specific terminology should be avoided here (or if it even can be), but for all isolates is a confusing name I think this is the best way to phrase this concisely. If that terminology should be avoided then I can describe what an isolate is here, though that feels a little out of scope for this section. If you think I should explain it please let me know and I can fix it.

cves/CVE-2015-1281.yml
Moremothansasuke: Sorry, I am kind of confused about what part you are talking about. If you mean the mistakes section I edited how the original author had worded things, but I also agree that this is the most important thing to note in the mistakes section. I don't feel like the mistakes section is where analysis of what the fix actually accomplished should go, if you think it should be added where do you think it should go?

PR #300, opened 2019-10-25
CVE-2016-5166 and CVE-2011-2782

Inchworm333:
👨‍💻	(╯°□°）╯︵ ┻━┻

### DIFF
--- a/cves/CVE-2011-2782.yml
+++ b/cves/CVE-2011-2782.yml
@@ -11,7 +11,7 @@ curated_instructions: |
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
--- a/cves/CVE-2011-2782.yml
+++ b/cves/CVE-2011-2782.yml
@@ -32,9 +32,10 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: The drag-and-drop implementation in Google Chrome on Linux does not properly
-  enforce permissions for files, which allows malicious actors to get around potential
-  access restrictions on files
+description: | The drag-and-drop implementation in Google Chrome on Linux does not properly
+  enforce permissions for files. This allows attackers to create a file link on their 
+  website (e.x. <a href="file:///etc/passwd">) and if the user drags this link onto something
+  like a file upload, and it will upload the file from the users device.
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
@@ -161,7 +162,7 @@ lessons:
   least_privilege:
     applies: true
     note: only use the user's normal privileges when it comes to loading files instead
-      of giving them greater privilege
+      of giving them greater privilege in order to access the file
   frameworks_are_optional:
     applies: 
     note: 
@@ -171,7 +172,7 @@ lessons:
   distrust_input:
     applies: true
     note: you should not assume that the user has permissions to view the file they
-      drag in and check to ensure they do
+      drag in and check to ensure they do.
   security_by_obscurity:
     applies: 
     note: 
@@ -204,4 +205,4 @@ mistakes:
     engineering industry would find interesting.
   answer: This vulnerability was clearly a design flaw as the developers, from the
     conversation on the bug, clearly did not anticipate drag-and-drop not working
-    properly when a URL linking to a file was dragged in as opposed to a file
+    properly when a URL linking to a file was dragged in as opposed to the file itself.
--- a/cves/CVE-2011-2782.yml
+++ b/cves/CVE-2011-2782.yml
@@ -32,7 +32,8 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: | The drag-and-drop implementation in Google Chrome on Linux does not properly
+description: | 
+  The drag-and-drop implementation in Google Chrome on Linux does not properly
   enforce permissions for files. This allows attackers to create a file link on their 
   website (e.x. <a href="file:///etc/passwd">) and if the user drags this link onto something
   like a file upload, and it will upload the file from the users device.
--- a/cves/CVE-2016-5166.yml
+++ b/cves/CVE-2016-5166.yml
@@ -11,7 +11,7 @@ curated_instructions: |
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
--- a/cves/CVE-2016-5166.yml
+++ b/cves/CVE-2016-5166.yml
@@ -127,7 +127,7 @@ interesting_commits:
   - commit: eff8e457298d01b437e4fd78194ad6de3c8d7ad6
     note: |
     The interesting part about this commit is the discussion on the code review page 
-    this (https://codereview.chromium.org/2075273002). During this code review they are initially trying 
+    "https://codereview.chromium.org/2075273002". During this code review they are initially trying 
     to figure out what external entity can access / trigger the save page dialog in order to make this a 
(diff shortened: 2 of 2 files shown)


cves/CVE-2011-2782.yml
```diff
@@ -11,7 +11,7 @@ curated_instructions: |
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
```
andymeneely: Looks like this is one that previously curated. Make sure you improve upon this one - don't just leave it there.

cves/CVE-2016-5166.yml
```diff
@@ -32,7 +32,7 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 
+description: Using the 'save page' feature that saves the web-page and web resources also saves locally uploaded files to that page.
```
CatherineOsadciw: Expand upon this description: what could an attacker do with this functionality? Does this allow them to access user information they shouldn't be able to?

cves/CVE-2016-5166.yml
```diff
@@ -198,4 +203,6 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: | 
+    This seems to have been a case where at the start this wasn't a vulnerability but as other technology got layered on top of it like chrome extensions and web-workers it became a
+    potential vulnerability from using this newer technology to exploit older ideas.
```
CatherineOsadciw: Expand upon your analysis: was this a coding mistake, a design mistake, a reqs mistake, or something else entirely?  Was there anything the devs could have potentially done to prevent the vulnerability from occurring in the first place?  Does the fix appear sufficient? etc.

cves/CVE-2016-5166.yml
```diff
@@ -32,7 +32,7 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description:
```
dis446: Small note, could you limit each line to something like 80 characters. It can get cumbersome to keep having to scroll horizontally on some editors

cves/CVE-2016-5166.yml
```diff
@@ -55,8 +55,8 @@ fixes:
 - :commit: eff8e457298d01b437e4fd78194ad6de3c8d7ad6
   :note: ''
 vccs:
-- :commit:
-  :note: 
+- :commit: 09911bf300f1a419907a9412154760efd0b7abc3
+  :note: 'The issue technically was created before they added chrome to the version control so technically it existed before this initial commit.'
```
dis446: Is there any indication as to when the vulnerability was introduced? Has it been there since the start?

cves/CVE-2016-5166.yml
```diff
@@ -107,7 +107,7 @@ subsystem:
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
-  answer: 
+  answer: "download"
   name:
```
dis446: i think "download" is meant to go under name

cves/CVE-2016-5166.yml
```diff
@@ -154,37 +159,37 @@ lessons:
...
     free to give it a small name and add one in the same format as these.
   defense_in_depth:
-    applies: 
+    applies: false
     note: 
   least_privilege:
-    applies: 
-    note: 
+    applies: true
+    note: Just because an attacker shouldn't be able to call save page doesn't mean they can't.
```
dis446: This doesn't apply to least privilege. Least privilege means users should only be allowed to do what they absolutely need to do, no more. This is a complexly separate statement

cves/CVE-2016-5166.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: https://cwe.mitre.org/data/definitions/616.html
```
KevinWongRIT: This CWE is about attackers overwriting global variables while CVE-2016-5166 is about the "Save page as" feature missing restrictions.  Is there another CWE that you think fits this situation better?

cves/CVE-2016-5166.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
...
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
```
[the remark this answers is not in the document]
KevinWongRIT: One of the example YAMLs given included the time while the other didn't so I'm not sure if either way is preferred.

cves/CVE-2016-5166.yml
```diff
@@ -95,10 +95,10 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
-  date: 
-  automated: 
-  google: 
+  answer: https://bugs.chromium.org/p/chromium/issues/detail?id=616429
```
KevinWongRIT: The answer should be in longform rather than just a link.

cves/CVE-2016-5166.yml
KevinWongRIT: Also, the examples didn't have quotes around the word.  It shouldn't matter unless the professor wanted to run a script to get all the CVEs for a subsystem.

cves/CVE-2016-5166.yml
```diff
@@ -118,10 +118,12 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer:
+  answer: |
```
KevinWongRIT: This explanation should be under the "note:" section after the commit.

cves/CVE-2016-5166.yml
```diff
@@ -55,8 +55,8 @@ fixes:
 - :commit: eff8e457298d01b437e4fd78194ad6de3c8d7ad6
   :note: ''
 vccs:
-- :commit:
-  :note: 
+- :commit: 09911bf300f1a419907a9412154760efd0b7abc3
+  :note: 'The issue technically was created before they added chrome to the version control so technically it existed before this initial commit.'
 upvotes_instructions: |
```
KevinWongRIT: +6

cves/CVE-2016-5166.yml
Inchworm333: As far as I know it has been there from the beginning. I can't think of any other places to look so I am assuming it has been there since the start.

cves/CVE-2016-5166.yml
```diff
@@ -118,10 +118,12 @@ interesting_commits:
     emerging themes?
 
     If there are no interesting commits, demonstrate that you completed this section by explaining what happened between the VCCs and the fix.
-  answer:
+  answer: | 
+    The interesting part about this commit is the discussion on the code review page (https://codereview.chromium.org/2075273002). During this code review they are initially trying
```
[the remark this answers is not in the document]
Inchworm333: Yes, all of the other specific commits were mostly just spelling changes so nothing interesting. I had no idea where I should put this though.

cves/CVE-2016-5166.yml
Inchworm333: The problem is that this seems like such a specific problem that there doesn't seem to be any suitable CWE that applies to it.

cves/CVE-2016-5166.yml
Inchworm333: The attacker can't really do anything with this functionality because it saves the documents to the users own computer. The developers were having trouble finding out how this could be used maliciously as well until they came up with the idea that maybe it could be abused by chrome extensions.

cves/CVE-2016-5166.yml
Inchworm333: I am not sure what you mean by longform.

cves/CVE-2011-2782.yml
Inchworm333: :(

cves/CVE-2016-5166.yml
Inchworm333: It started out as just a normal feature but later on in the programs life, due to expanding functionality it became a vulnerability later in time and would probably now be considered a design mistake.

PR #302, opened 2019-10-25
CVE-2011-1196

exw4141:


cves/CVE-2011-1196.yml
```diff
@@ -1,17 +1,16 @@
----
 CVE: CVE-2011-1196
 CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 122
+CWE: 119
```
vince21298: Can you discuss the reason(s) why you switch from CWE 122 to CWE 119?  In my opinion, CWE 122 is more closely related to this vulnerability since it was caused by a heap overflow.

cves/CVE-2011-1196.yml
```diff
@@ -32,11 +31,11 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: "Heap-based overflow in Google Chrome's implementation of the OGG metafile
-  \nformat allowed remote attackers to cause a denial of service or possibly have
-  \nunspecified other impact via unknown vectors that trigger an out-of-bounds \nwrite.\n\nThe
-  fix was initializing a variable and validating the indexes within the \nOGG stream
-  buffer.\n"
+description: |
+  The OGG container allows for the streaming and presenting of digital multimedia to users. A OGG
```
vince21298: I am still not sure what OGG is. Can you explain a little more in detail about what OGG is?

cves/CVE-2011-1196.yml
```diff
@@ -59,20 +58,17 @@ fixes_vcc_instructions: |
...
-    \nOGG stream buffer.\n"
-- :commit: 4ff6ce9702a91c52892786464b15f3c8588d4070
-  :note: 'Bump the version number of the patched ffmpeg dependency
-
-'
-- :commit: e9e0a7d9f8fe4a3faecd62091a1910b1b27488cf
-  :note: 'Bump the version number of the patched ffmpeg dependency
-
-'
+- commit:
```
vince21298: Try greping in the git log with the bug number, etc. I found this commit `e9e0a7d9f8fe4a3faecd62091a1910b1b27488cf`. You should check it out and see if this the final fix commit. Additionally, you might be able to find the VCC now.

cves/CVE-2011-1196.yml
```diff
@@ -59,20 +58,17 @@ fixes_vcc_instructions: |
...
+    indexes within the nOGG stream buffer.
+- commit: 4ff6ce9702a91c52892786464b15f3c8588d4070
+  note: Bump the version number of the patched ffmpeg dependency
+- commit: e9e0a7d9f8fe4a3faecd62091a1910b1b27488cf
+  note: Bump the version number of the patched ffmpeg dependency
 vccs:
-- :commit:
-  :note: 
+- commit:  
+  note:
```
Bamagik: There should be a commit that was the cause of the issue

cves/CVE-2011-1196.yml
```diff
@@ -59,20 +58,17 @@ fixes_vcc_instructions: |
...
-
-'
+- commit: 
+  note: |
+    Revision 76146 in Chromium. Unable to find in Github. The fix was initializing a variable and validating the 
+    indexes within the nOGG stream buffer.
+- commit: 4ff6ce9702a91c52892786464b15f3c8588d4070
+  note: Bump the version number of the patched ffmpeg dependency
+- commit: e9e0a7d9f8fe4a3faecd62091a1910b1b27488cf
+  note: Bump the version number of the patched ffmpeg dependency
```
Bamagik: I'm unsure if version number bumping needs to be represented as part of the fix commits

cves/CVE-2011-1196.yml
Bamagik: .OGG is just a file format for audio/video. I dont think it needs too much explanation. 

Although this could use a more detailed explanation of why the overflows occur within the codec on reading

Bamagik:
4 Votes added

cves/CVE-2011-1196.yml
vince21298: I tried looking into it and I couldn't seem to find the VCC. I believe I saw an SVN revision number. I think that number should be placed there instead of nothing.

vince21298:
2 Votes Added

cves/CVE-2011-1196.yml
exw4141: I was thinking the same thing. I just saw that the NVD stated 119 so I just took their word for it. I'll switch it back.

cves/CVE-2011-1196.yml
exw4141: A little more detail was added illustrating how a user may encounter it

cves/CVE-2011-1196.yml
exw4141: I have removed those commits.

cves/CVE-2011-1196.yml
exw4141: Interesting. I did not even find a SVN revision number. I did find a supposed link to their changelog but it did not work since it has been 8 years since it was posted.

exw4141: Found a couple of related revisions in Chromium

PR #305, opened 2019-10-25
Adding to CVE-2017-5110

cxd3796:


JMdoubleU:
Do you think that CWE 346 or 451 would be more appropriate? CWE 290 is related to spoofing authentication schemes, which doesn't seem to be part of this issue.

Were you unable to find a VCC for this? It's currently not defined in your file.

It might be helpful to make the description a bit less chromium-specific (rework reference to `PaymentRequestAPI .show()`) and perhaps elaborate on what "origin" means in this context.

cves/CVE-2017-5110.yml
```diff
@@ -98,20 +98,20 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: The vulnerability was stumbled upon as other general features were being explored.
```
Rxy8841: By "stumble upon " does that mean during code review or unit testing, or in rare cases a developer realizing and seeing there was a problem?

cves/CVE-2017-5110.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
...
+CWE: 290
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
```
Rxy8841: I don't fully understand the vulnerability. Was the function show() open to users, and they were able to change values when they were not suppose to? Also what do you mean by origin payment.

cves/CVE-2017-5110.yml
```diff
@@ -157,38 +157,38 @@ lessons:
     If you think of another lesson we covered in class that applies here, feel
     free to give it a small name and add one in the same format as these.
   defense_in_depth:
-    applies: 
-    note: 
+    applies: false
+    note:
   least_privilege:
```
Rxy8841: Would least privileges apply in this case because the application is allowing a attacker or user access a function that they should not be able to?

cves/CVE-2017-5110.yml
```diff
@@ -157,38 +157,38 @@ lessons:
...
   environment_variables:
-    applies: 
-    note: 
+    applies: false
+    note:
   secure_by_default:
-    applies: 
-    note: 
+    applies: true
+    note: Believing that an application is secure by default can lead a developer to not properly architect or design their solution with all requirements in mind. This may have happened somewhere along the way.
```
Rxy8841: It would be nice if you use the "|" symbol and keep the words on the page so readers don't have to scroll. But not a big deal.

Rxy8841: I also agree with the principle being applied here!!! It could possibly be the source of the problem.

cves/CVE-2017-5110.yml
```diff
@@ -80,9 +80,9 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: There appeared to be some tests written in the code, whether or not they were utilized was a different story though.
```
JosefLenz: Its not good to leave ambiguity, when you say "whether or not they were utilized", were the tests there for the vulnerability and not working correctly? Did the tests not cover the possibility of this vulnerability occurring? Or were the tests just not being run?

cves/CVE-2017-5110.yml
```diff
@@ -80,9 +80,9 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: There appeared to be some tests written in the code, whether or not they were utilized was a different story though.
-  code: There was code written to help automate the testing process.
-  fix: The fix did involve improving tests alongside other code.
+  answer: There appeared to be some tests written in the code, whether or not they were utilized was a different story though. There was code written to help automate the testing process. The fix did involve improving tests alongside other code.
```
JosefLenz: It would be good to break this to multiple lines to improve readability, you can specify a multi line answer with the "|" character, I saw a few places where this could help improve your doc.

cxd3796: Yup, I figured this out afterwards, thanks!

cves/CVE-2017-5110.yml
cxd3796: Will edit.

cves/CVE-2017-5110.yml
cxd3796: Least privilege doesn't apply here, because the problem at hand is spoofing. It doesn't matter how much privilege is being given out if someone can spoof the system's origin point, allowing for a middle man. I do agree that least privilege should be followed, but it doesn't pertain to this vulnerability.

PR #309, opened 2019-10-25
CVE-2017-5109

iclare:
Update cves/CVE-2017-5109.yml

### DIFF
--- a/cves/CVE-2017-5109.yml
+++ b/cves/CVE-2017-5109.yml
@@ -4,14 +4,14 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: 20
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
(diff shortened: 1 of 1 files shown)


iclare:
> You identified this as an instance of "CWE-20: Improper Input Validation". While I think you are right that additional input validation may have been a possible solution to this problem, have you considered "CWE-1021: Improper Restriction of Rendered UI Layers or Frames?" This seems to fit the description of this vulnerability a bit better, since it is a matter of "frame objects or UI layers that belong to another application or domain" being rendered while accessing a web page.
> 
> Also, your description may benefit from mentioning something about the "unload handler handling" that is mentioned in the CVE. This seems like it could be a relevant detail about the vulnerability - if for nothing else, at least to give the reader some background knowledge.

Thanks for that CWE, I didn't see that one. I'm not entirely sure if this one applies.

This is the description for CWE-1021.

>The web application does not restrict or incorrectly restricts **frame objects or UI layers** that belong to another application or domain, which can lead to user confusion about which interface the user is interacting with."

The permission request that pops up is a part of Chrome's own UI. I'm not sure what's meant by _belongs to_. It it means to say that the context belongs to another application or domain, then I think CWE-1021 does apply to CVE-2017-5109.

exl5325:
1 upvote for this vulnerability

cves/CVE-2017-5109.yml
```diff
@@ -98,20 +103,25 @@ discovered:
...
 
     Look at the path of the source code files code that were fixed to get
     directory names. Look at comments in the code. Look at the bug reports how
     the bug report was tagged. Examples: "clipboard", "gpu", "ssl", "speech", "renderer"
-  answer: 
-  name: 
+  answer: |
+    The bug report component is tagged as UI>Browser>Permissions>Prompts.
+    The VCC is found in chromium/content/browser/permissions.
+  name: chromium/content/browser/permissions
```
CatherineOsadciw: Name should be something like "permissions" rather than the full filepath (clogs up the VHP subsystem tags otherwise)

etiano417:
I would like to give 3 upvotes to this vulnerability

rburchlau001:
3 upvotes here

PR #312, opened 2019-10-25
CVE-2017-5052

wrs1492:
Vulnerability in the rendering engine that caused memory allocation issues when blocks were nested in each other.

### DIFF
--- a/cves/CVE-2017-5052.yml
+++ b/cves/CVE-2017-5052.yml
@@ -81,7 +81,13 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: There were automated unit tests in this vulnerability and the original
+  answer: |
+    There were automated unit tests in this vulnerability and the original
+    code was unit tested. The fix did not involve improving the automated tests. 
+  code: |
+    The bug was an extremely niche situation, so while the code was unit tested,
+    it did not get the use case that caused the vulnerability.
+    fix: A unit test was added for this issue. 
   code was unit tested. The fix did not involve improving the automated tests. 
   code: The bug was an extremely niche situation, so while the code was unit tested,
   it did not get the use case that caused the vulnerability.
--- a/cves/CVE-2017-5052.yml
+++ b/cves/CVE-2017-5052.yml
@@ -11,7 +11,7 @@ curated_instructions: |
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: True
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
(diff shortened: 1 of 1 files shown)


cves/CVE-2017-5052.yml
```diff
@@ -80,9 +81,11 @@ unit_tested:
 
     For the "fix" answer below, check if the fix for the vulnerability involves
     adding or improving an automated test to ensure this doesn't happen again.
-  answer: 
-  code: 
-  fix: 
+  answer: There were automated unit tests in this vulnerability and the original
```
mesternefeld: This should help fix the CI
```suggestion
  answer: |
    There were automated unit tests in this vulnerability and the original
    code was unit tested. The fix did not involve improving the automated tests. 
  code: |
    The bug was an extremely niche situation, so while the code was unit tested,
    it did not get the use case that caused the vulnerability.
    fix: A unit test was added for this issue. 
```

cves/CVE-2017-5052.yml
```diff
@@ -201,4 +206,9 @@ mistakes:
     Use those questions to inspire your answer. Don't feel obligated to answer
     every one. Write a thoughtful entry here that those ing the software
     engineering industry would find interesting.
-  answer: 
+  answer: It is hard to say what type of mistake this was as it was such a niche
+  vulnerability on a wierd edge case that ended up having very serious security
+  implications. I would have to say this falls under a coding mistake as the
```
bjt6087: "weird edge case"

bjt6087:
Would CWE-119 fit this vulnerability?

bjt6087:
Interesting vulnerability. 3 upvotes

cves/CVE-2017-5052.yml
```diff
@@ -98,20 +101,22 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
-  date: 
-  automated: 
-  google: 
-  contest: 
+  answer: The vulnerability was discoverd by a misuse case, which caused chrome to crash
```
Ramenous: If possible, can you describe and elaborate on what was the misuse case that caused chrome to crash? Providing the proper use case along with the misuse case can also help differentiate what was the difference in flow that caused this vulnerability

cves/CVE-2017-5052.yml
TLezynski: Are you certain it was a coding mistake? If this is an edge case, perhaps the edge case wasn't considered in the design and it's a design flaw.

cves/CVE-2017-5052.yml
Ramenous: If possible, can you describe and elaborate on what was the misuse case that caused chrome to crash? Providing the proper use case along with the misuse case can also help differentiate what was the difference in flow that caused this vulnerability

cves/CVE-2017-5052.yml
wrs1492: I would not say it was a design flaw as the fix for the bug was a relatively simple change, and took care of the problem thoroughly. Going through the history of the bug, it was really hard to produce a test case that would reliably cause the bug to occur, so I would say it is more of a testing mistake than a coding one, so I will update the file accordingly

cves/CVE-2017-5052.yml
wrs1492: Yeah sorry I ran out of time when submitting this vulnerability, the crash was caused by nested "blocks", and I will flesh out this answer.

wrs1492:
> line 7: I see that you didn't add a CWE for this vulnerability. Is that intentional in that you couldn't find an appropriate one? or did you miss it?

I ran out of time when submitting, I will find an appropriate one

wrs1492:
> CWE-119

I think this might be a good fit, I will see if there are any others that fit better. Thanks for the help.

rlb8800 (approved):

rlb8800:
Fixed this on dev

PR #313, opened 2019-10-25
CVE-2015-1224

AdamVD:


ath9902:
I thought this vulnerability was very interesting because the developer made assumptions about th data sizes which weren't necessarily true. I also think CWE-120: Buffer Copy without Checking Size of Input could also apply here. I want to give +2 upvotes.

AdamVD:
@ath9902 Thanks for the comment and +2 (upvotes have been updated). CWE-120 is close to the issue at hand. The difference is that rather than trying to place too much data in the copy destination, too much data was being read from the source. Upon further consideration prompted by your comment, the CWE number has been changed to 125 - an out-of-bounds read.

cves/CVE-2015-1224.yml
```diff
@@ -4,14 +4,14 @@ CWE_instructions: |
...
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
```
CatherineOsadciw: Remove time portion on the "announced" field below

CatherineOsadciw: + bounty date below

cves/CVE-2015-1224.yml
```diff
@@ -216,7 +216,19 @@ mistakes:
     \ Initially,\nthey did have support for video streams with alpha buffers larger
     than the image \nbuffer, but the library changed a year after the first implementation
     and\nintroduced this vulnerability.  They didn't account for that and this vulnerability\nwas
-    introduced.\n\nTo fix this, they added a check that threw an error if the alpha
-    buffer was larger\nthan the image buffer.  These buffers are heap buffers, so
+    introduced.
+    
+    \n\nThe code written to copy video data from the buffer assumed that the image and alpha plane
```
CatherineOsadciw: You can probably delete these newlines

cves/CVE-2015-1224.yml
AdamVD: Those were filled in by the bots but I will remove for consistency and safety. Thanks!

cves/CVE-2015-1224.yml
AdamVD: It wasn't clear how to format this data initially, but @andymeneely told me recently that Markdown formatting will work. Although I'm not sure what the syntax needs to be, I've removed any '\n's and converted my paragraphs into single lines.

aaa1008:
Do you have any insight on why the previous CWE number was 122 (CWE-122: Heap-based Buffer Overflow)?

AdamVD:
@aaa1008 I think it was a slight misunderstanding of the issue. As I wrote before:
> The difference is that rather than trying to place too much data in the copy destination, too much data was being read from the source. Upon further consideration prompted by your comment, the CWE number has been changed to 125 - an out-of-bounds read.

If the issue was an overflow during a *write to the copy destination*, it would have been CWE 122.

PR #314, opened 2019-10-25
CVE-2010-0656 and CVE-2013-6657

NBrockman:
Finished CVE-2010-0656 and added curated to CVE-2013-6657, done

### DIFF
--- a/cves/CVE-2010-0656.yml
+++ b/cves/CVE-2010-0656.yml
@@ -4,21 +4,21 @@ CWE_instructions: |
   Please go to cwe.mitre.org and find the most specific, appropriate CWE entry
   that describes your vulnerability. (Tip: this may not be a good one to start
   with - spend time understanding this vulnerability before making your choice!)
-CWE: 
+CWE: 20
 curated_instructions: |
   If you are manually editing this file, then you are "curating" it. Set the
   entry below to "true" as soon as you start. This will enable additional
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
   source for this is Chrome's Stable Release Channel
   (https://chromereleases.googleblog.com/).
   Please enter your date in YYYY-MM-DD format.
-announced: 2010-02-18 13:00:00.847000000 -05:00
+announced: 2010-01-25
 description_instructions: |
   You can get an initial description from the CVE entry on cve.mitre.org. These
   descriptions are a fine start, but they can be kind of jargony.
@@ -32,10 +32,11 @@ description_instructions: |
   that outsiders to Chromium would not understand. Technology like "regular
   expressions" is fine, and security phrases like "invalid write" are fine to
   keep too.
-description: 'Webkit shows a page that displays directories on the server in response
-  to an XMLHttpRequest for a file that''s actually a directory.
+description: |
+  Webkit shows a page that displays directories on the server in response
+  to an XMLHttpRequest for a file that's actually a directory. Based upon the XHR
+  requests to chrome browser.
 
-'
 bounty_instructions: |
   If you came across any indications that a bounty was paid out for this
   vulnerability, fill it out here. Or correct it if the information already here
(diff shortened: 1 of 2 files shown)


cves/CVE-2010-0656.yml
```diff
@@ -167,8 +173,11 @@ lessons:
     note: 
   least_privilege:
     applies: true
-    note: Anyone was able to see the server directory, when they definitely should
-      not be able to.
+    note: |
+      Anyone was able to see the server directory, when they definitely should
+      not be able to. As abarth@chromium.org said, "It's better do deny access to directory listings. 
+      This makes it harder for malicious local HTML pages to spider 
+      your file system.""
```
Jeremy523: I think I know what you're trying to say when you say "spider your file system", but I'd try to word it a little bit clearer way.

cves/CVE-2010-0656.yml
```diff
@@ -188,8 +197,10 @@ lessons:
     applies: 
     note: 
   secure_by_default:
-    applies: 
-    note: 
+    applies: true
+    note: |
+      This concept applies to why this bug was happening to Chrome but not other browsers.
+      Other browsers made a secur by default policy that kept them safe but less flexible/modular.
```
Jeremy523: Not clear on what you're trying to say with your first sentence and how it relates to the concept of "secure by default". How is Chrome's implementation not following the concept? Also check your spelling on line 203 with "secur".

cves/CVE-2013-6657.yml
```diff
@@ -11,7 +11,7 @@ curated_instructions: |
   integrity checks on this file to make sure you fill everything out properly.
   If you are a student, we cannot accept your work as finished unless curated is
   set to true.
-curated: false
+curated: true
 announced_instructions: |
   Was there a date that this vulnerability was announced to the world? You can
   find this in changelogs, blogs, bug reports, or perhaps the CVE date. A good
```
Jeremy523: Not certain if it's required but I'd change the `announced` field to just YYYY-MM-DD format without the time info.

cves/CVE-2013-6657.yml
```diff
@@ -105,7 +105,7 @@ discovered:
 
     If there is no evidence as to how this vulnerability was found, then you may
     leave the entries blank except for "answer". Write down where you looked in "answer".
-  answer: 
+  answer: Was found by a random chromium bug reporter.
```
Jeremy523: Can you expand on this? Was there much discussion in that report thread? I'd also recommend using complete sentences.

Jeremy523: Also in the below section `interesting_commits`, if you found no interesting commits, explain what happened between the VCCs and the fix.

Jeremy523: In the below section `major_events`, there are multiple `\n`'s likely from some improper formatting that should be fixed.

cves/CVE-2010-0656.yml
```diff
@@ -56,14 +57,17 @@ fixes_vcc_instructions: |
...
 fixes:
 - :commit: c0fbaed2be9b4df04c8cc9fe3f5c1c41b3459ca0
-  :note: ''
+  :note: 'Nothing test regressions'
 - :commit: 04c84bc6d9fdc15a8d49786b28ee2256aaaf50a8
-  :note: ''
+  :note: 'Showing the nonexistent file test failing'
 - :commit: 1c966c891f7f495fd08c84b58ec145bd6bd7d181
-  :note: ''
+  :note: 'Shows implementation of universal access to files for fileURL test fixes'
```
cxs7700: Complete sentences would make this easier to read and understand. For example, "nothing test regressions" is unclear to me.

cves/CVE-2010-0656.yml
```diff
@@ -56,14 +57,17 @@ fixes_vcc_instructions: |
...
 vccs:
 - :commit:
-  :note: 
+  :note: |
+    The origin of the bug is not really a commit or code error but a structural issue. Other browsers
+    such as Safari got away with not allowing local directory access by denying any file directory access 
+    from URLs.
```
cxs7700: You can probably do some more digging on the bug reports pages and the "Blame" page for each file to find the VCC.

cves/CVE-2010-0656.yml
```diff
@@ -188,8 +197,10 @@ lessons:
     applies:
```
cxs7700: Do you think distrust input could also be a lesson? Based on my understanding of your description, requests could have potentially been sanitized by the system to make sure they do not display on the browser. The description could also be more detailed for ease of understanding.

cves/CVE-2013-6657.yml
cxs7700: I was going to give the same feedback because the format stated in the prompt is enough. Many of these values seem like they were automatically generated, especially because this .yml file does not have many added changes. Make sure you double check each automatically generated result and improve on them.

cves/CVE-2013-6657.yml
cxs7700: For subsystems and where the mistake was in, it would help to either note the bug report or note down the path to the mistake.

cxs7700: What is "Blink"? Explaining what Blink is will help the reader better understand the system and vulnerability.

cves/CVE-2010-0656.yml
NBrockman: Spidering is the same as crawling or scoping out

cves/CVE-2010-0656.yml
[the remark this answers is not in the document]
NBrockman: The vulnerability is present because they use Webkit and the code that cause the file system to display is in the Webkit not Chrome so there is not a commit for the original VCC. There is though checks and tests that they add on their end to verify Webkits inputs to the chromium code and that is what I can give as the fixes.

cves/CVE-2010-0656.yml
NBrockman: Yes It could be, thank you for pointing it out. The input from Webkit is probably the easier and more important thing to check in this instance for the this vulnerability.

cves/CVE-2010-0656.yml
NBrockman: I have elaborated a bit on that now, thank you

iclare:
1 upvote - If I understand it correctly, seems like a good entry point for finding more vulnerabilities or info.

exl5325:
3 upvotes

PR #316, opened 2019-11-01
CVE-2011-3875

MaeveMonster:


cves/CVE-2011-3875.yml
```diff
@@ -158,38 +164,38 @@ lessons:
...
-    applies: 
-    note: 
+    applies:
+    note:
   native_wrappers:
-    applies: 
-    note: 
+    applies:
+    note:
   distrust_input:
```
Ramenous: I think this vulnerability would apply to distrust input as attackers could spoof URL bars through unspecified vectors

